Issue #175, opened 2026-06-16
Confirm the manifest shape and signed_fields are sufficient for a gateway to bind a runtime session (cMCP #302)


carloshvp:
## Spec section
Section 3.1 (root model: agent_id, issuer, artifacts) and Section 3.6 (signature, signed_fields), plus the verification API.

## Problem
cMCP issue agentrust-io/cmcp#302 (Option A, confirmed) will have the gateway ingest a signed Agent Manifest at session create and bind it to the runtime session: verify the issuer signature, check that the session SVID equals manifest.agent_id, and confirm the loaded policy-bundle and tool-catalog hashes equal the manifest's bound hashes. Before finalizing that gateway change I want to confirm the manifest shape already supports it, and pin the exact verification contract so the gateway does not reconstruct (and drift from) the canonical signing pre-image.

## Proposed change (confirmations, likely no normative change)
1. Confirm agent_id (SPIFFE), issuer (SPIFFE), and artifacts (policy_bundle.hash, the tool_manifest catalog hash, system_prompt.hash) are all in SIGNED_FIELDS and therefore authenticated by the manifest signature. They appear to be.
2. Confirm the gateway should verify via the SDK canonical verification entry point (which selects the SIGNED_FIELDS subset and canonicalizes per ADR-0001 RFC 8785) rather than rebuilding the pre-image itself. Please point to the public function to call.
3. Algorithm scope for the first cMCP iteration: is Ed25519 (standard profile) sufficient, or must the gateway also accept hybrid-Ed25519-ML-DSA-65 and ML-DSA-65 at launch?
4. Expiry: the gateway will reject a manifest when now is outside [issued_at, expires_at]. Confirm that is the intended behavior.
5. Delegation: when delegation_chain is present, which identity should the gateway bind the session SVID to, the leaf agent_id or each hop? Out of scope for the first iteration, but I want to record the intended direction.
6. The attestation field is appended by the TEE at launch and excluded from the signature pre-image. Confirm the gateway should ignore it for the agent_id binding (the gateway has its own session attestation) and not treat its absence as a failure.

## Impact
- Backward compatible: yes (no manifest format change expected).
- Affects conformance level(s): Level 1+ (binding).
- Conformance tests that need updating: none expected.
- Regulatory mapping impact: none.

## Alternatives considered
Gateway reconstructs the signing pre-image itself: rejected, it would risk canonicalization drift from the SDK. Reusing the SDK verifier is safer.

PR #176, opened 2026-06-17
feat(sdk): export the relying-party verification API from the package root

carloshvp:
## Summary
Export the relying-party verification engine from the package root, so
`agent_manifest.verify_manifest` is the canonical entry point a verifier or
gateway calls.

## Problem
`verify_manifest` and its result/context types live in the private
`agent_manifest._verify` module. They are documented in
`docs/api-reference/verification.md`, but are not in the package's public API
(`__all__`). Today relying parties, and even this repo's own tests, import them
from the private `_verify` path. In the cMCP integration (agentrust-io/cmcp#302)
this led to reimplementing the canonical signing pre-image by hand instead of
reusing the SDK, which is the concern raised in #175.

## Change
Export the hosting-agnostic engine and result types from `agent_manifest`:
`verify_manifest`, `VerificationContext`, `VerificationResult`, `OverallResult`,
`FieldResult`, `DelegationResult`, `HitlResult`, `FieldsVerified`,
`MismatchDetail`, `EvidencePack`, `RevocationStore`, `RevocationRecord`.

`create_router` (the FastAPI HTTP wiring) is intentionally left in the submodule,
so importing the package does not pull FastAPI as a hard dependency (FastAPI is
imported lazily inside `create_router`).

## Tests
`tests/test_public_api.py`: asserts the API is importable from the package root,
plus a VALID-then-MISMATCH roundtrip through the public path.

## Validation
- `pytest tests`: 440 passed, 22 skipped (no regression)
- `ruff check`: clean
- `import agent_manifest` does not require FastAPI

Non-breaking addition; existing imports are unaffected.

## Open question for maintainers
This exposes the engine that currently lives under `_verify`. Since it is already
documented as public API, exporting it at the top level seems right, but if you
prefer a different public surface (for example a dedicated `agent_manifest.verify`
namespace), happy to adjust. This PR is the concrete answer to #175.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to Agent Manifest are documented here. Format follows [Keep 
 
 ## [Unreleased]
 
+### Added
+
+**[SDK]** Export the verification API from the package root, so relying parties
+and gateways call `agent_manifest.verify_manifest()` and `VerificationContext`
+directly instead of importing the private `_verify` module (#175).
+
 ## [0.1.0] — 2026-06-23
 
 Stable launch release at Confidential Computing Summit, June 23 2026.
--- a/docs/api-reference/verification.md
+++ b/docs/api-reference/verification.md
@@ -2,6 +2,24 @@
 
 The core verification engine and FastAPI router. See [Tutorial: Server-side verification](../tutorials/server-side-verification.md) for usage examples.
 
+## Public API
+
+For gateway and runtime-session binding, call the package-root export rather than
+the private `_verify` module:
+
+```python
+from agent_manifest import RevocationStore, VerificationContext, verify_manifest
+```
+
+`verify_manifest()` is the supported high-level entry point.
+`VerificationContext.trusted_keys` maps an issuer `key_id` (the SHA-256 hex of
+the public key bytes) to its base64url-encoded Ed25519 public key, the form
+returned by `Ed25519KeyPair.public_b64url()`. A consumer that holds raw public
+key bytes must base64url-encode them before populating `trusted_keys`. Signers
+and verifiers share `agent_manifest.signing_pre_image()` for the exact RFC 8785
+canonical byte sequence, including the `hitl_record.approvals` normalization, so
+a relying party never reconstructs the pre-image itself.
+
 ## Core function
 
 ::: agent_manifest._verify.verify_manifest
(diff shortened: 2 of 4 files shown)


imran-siddique (commented):
Thanks for this - making the verification engine a stable public API is the right call and the test showing a VALID→MISMATCH roundtrip through the new import path is exactly the right validation. A few things to work through:

**`docs/api-reference/verification.md` not updated**

The PR description notes the verification types "are documented in `docs/api-reference/verification.md`" but the doc is not updated in this PR. If the doc currently points users to `agent_manifest._verify`, it needs to be updated to `agent_manifest` and the private import path deprecated. Exporting from `__init__` is the right fix but it lands half-finished if the docs still show the old path.

**`EvidencePack`, `RevocationStore`, `RevocationRecord` need public documentation**

These three types are added to `__all__`. They are less obviously required for basic verification than `verify_manifest` and `VerificationResult`, so a user landing on the API for the first time will not know when to use them. Add at least a one-sentence description to `docs/api-reference/verification.md` for each (or to the module docstring).

**`VerificationContext.trusted_keys` type mismatch with cMCP**

The test calls `keypair.public_b64url()` as the value in `trusted_keys`, producing `dict[str, str]` (base64url-encoded key string). But cMCP PR #315, which is the stated consumer of this export, uses `trusted_agent_manifest_keys: dict[str, bytes]` (raw public key bytes). These type signatures do not match. Before the cMCP integration can call `agent_manifest.verify_manifest` rather than reimplementing the pre-image, the type needs to align. Document the expected format in the public API, or align the types between the two packages.

**No end-to-end cMCP integration test**

The PR's motivation is that cMCP PR #315 reimplemented the signing pre-image by hand instead of calling this SDK. Exporting the API here solves half the problem; the other half is that cMCP needs to be updated to call `agent_manifest.verify_manifest` rather than its custom `agent_manifest.py`. As written, both implementations exist simultaneously - the SDK and the cMCP reimplementation. Add a note or a follow-up issue tracking that cMCP should be updated to use this export, and confirm the types are compatible before that update lands.

**Test: empty `RevocationStore` and partial `VerificationContext`**

The test creates a full `VerificationContext` with all artifact fields matching the manifest. What happens when `VerificationContext` has some fields but the manifest has artifacts the context does not cover? Is that `VALID` (no assertion about unchecked fields) or something else? A test for partial context would clarify the intended semantics.

carloshvp:
Addressed the review in `59010d2`:

- `docs/api-reference/verification.md` now points relying parties to the package-root export and documents the `trusted_keys` format (issuer `key_id` to base64url-encoded Ed25519 public key; a consumer holding raw bytes must encode them first), plus one-line descriptions for `EvidencePack`, `RevocationStore`, and `RevocationRecord`.
- added a partial-`VerificationContext` test (unprovided fields are `NOT_BOUND`, the result stays `VALID`).
- CHANGELOG notes the export.

On the cMCP integration point: the spec contract is split out into #177, and the follow-up to migrate the cMCP gateway off its in-repo reimplementation to call `agent_manifest.verify_manifest` (and to align the `trusted_keys` type) is tracked in agentrust-io/cmcp#317.

PR #177, opened 2026-06-17
docs: define runtime manifest binding contract

### LINKED ISSUE #175: Confirm the manifest shape and signed_fields are sufficient for a gateway to bind a runtime session (cMCP #302)

carloshvp:
## Summary
Add normative runtime-session binding guidance to the Agent Manifest spec for gateways such as cMCP: the signed fields that bind `agent_id` and the artifact hashes, validity-window checks, delegation handling, and the separation of attestation from the signed pre-image.

This is the spec-contract half of the work. The package-root export of the verification API (so gateways can call `agent_manifest.verify_manifest()`) is handled in #176, to keep the SDK change and the spec change reviewable separately.

Closes #175.

## Validation
- `python -m pytest` (SDK unchanged on this branch)
- `ruff check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to Agent Manifest are documented here. Format follows [Keep 
 
 ## [Unreleased]
 
+### Added
+
+**[SPEC]** Document runtime-session binding guidance for gateways, including
+the signed fields that bind `agent_id`, artifact hashes, validity windows,
+delegation handling, and attestation separation (#175).
+
 ## [0.1.0] — 2026-06-23
 
 Stable launch release at Confidential Computing Summit, June 23 2026.
--- a/spec/agent-manifest-spec-v0.1.md
+++ b/spec/agent-manifest-spec-v0.1.md
@@ -1055,6 +1055,60 @@ A `VALID` result means all of the following are true:
 
 A `MISMATCH` result means at least one required field does not match its running artifact. The `mismatch_details` array MUST enumerate every mismatched field. A relying party receiving a `MISMATCH` MUST NOT proceed with the operation that triggered verification.
 
+#### 5.3.1 Runtime session binding for gateways
+
+A gateway that binds an Agent Manifest to a runtime session, such as cMCP, MUST
+use the manifest verification API rather than reconstructing a signing
+pre-image locally. In the Python SDK, the public entry point is
+`agent_manifest.verify_manifest(manifest, context, revocation_store)`;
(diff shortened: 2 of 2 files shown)


imran-siddique:
Superseded by #178. The branch was rebased to resolve the CHANGELOG conflict with #176 and re-opened from the in-repo copy of the branch to unblock merge.

PR #220, opened 2026-07-06
fix(verify): bind trusted keys to manifest issuers

### LINKED ISSUE #215: security: bind trusted keys to issuers (kid->issuer authorization model)
Tracked follow-up from the June 2026 org-wide security hardening review. Wave 1/2 fixes are merged and published; this is remaining hardening.

verify_manifest checks that a signature's key_id exists in trusted_keys but not that the key is authorized for the manifest's issuer. In a multi-issuer trusted_keys map, issuer A could mint a manifest claiming issuer B and verify under A's key. Add an issuer<->key authorization binding and require the resolved key be authorized for the manifest issuer.

carloshvp:
## Summary
- add `trusted_key_issuers` to verifier contexts as a key_id -> issuer authorization map
- fail closed when the resolved signing key is not authorized for `manifest.issuer`
- thread the authorization map through `POST /verify` and cover cross-issuer regression cases

## Why
`verify_manifest` previously checked that `signature.key_id` existed in `trusted_keys`, but not that the trusted key was authorized for the issuer claimed by the manifest. In multi-issuer trust maps, that allowed a key trusted for issuer A to sign a manifest claiming issuer B.

Closes #215.

## Validation
- `python -m pytest tests/test_verify.py tests/test_am_verify.py -q`
- `python -m pytest -q`
- `pytest -v --tb=short --cov=agent_manifest --cov-report=term-missing --cov-fail-under=80`
- `ruff check src/ tests/ --select E,F,W --ignore E501`
- `mypy src/agent_manifest`
- `bandit -r src/agent_manifest -c pyproject.toml`
- `pip-audit`

### DIFF
--- a/python/src/agent_manifest/_verify.py
+++ b/python/src/agent_manifest/_verify.py
@@ -133,6 +133,10 @@ class VerifyRequest(BaseModel):
     the base64url-encoded public key. Signature verification is fail-closed:
     without trusted keys, a signed manifest yields ``UNVERIFIABLE`` and an
     unsigned manifest yields ``SIGNATURE_MISSING`` - never ``VALID``.
+
+    ``trusted_key_issuers`` maps each trusted key_id to the issuer SPIFFE URIs
+    authorized to sign manifests with that key. When supplied, key-to-issuer
+    authorization is fail-closed.
     """
 
     manifest_id: str
@@ -140,6 +144,8 @@ class VerifyRequest(BaseModel):
     enforce_attestation: bool = False
     # key_id (sha256 hex of pub key bytes) -> base64url-encoded public key bytes
     trusted_keys: dict[str, str] = Field(default_factory=dict)
+    # key_id -> issuer SPIFFE URIs authorized to use that signing key
+    trusted_key_issuers: dict[str, list[str]] = Field(default_factory=dict)
     # principal_id -> base64url-encoded public key bytes (for delegation chain)
     delegation_public_keys: dict[str, str] = Field(default_factory=dict)
     # When True, a manifest without a delegation_chain is a verification failure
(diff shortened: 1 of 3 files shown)


imran-siddique (approved):
Approving. Correct fail-closed fix for the key-to-issuer confused-deputy gap.

What I verified:
- `_signature_key_issuer_mismatch` (`python/src/agent_manifest/_verify.py`) is evaluated in the `else` branch after the `key_id in trusted_keys` check and BEFORE signature verification, so an unauthorized key never even reaches the crypto path. When `trusted_key_issuers` is supplied it fails closed on every unhappy case:
  - missing/empty `manifest.issuer` -> `signature.issuer` mismatch;
  - `key_id` absent from the map (`allowed_issuers` falsy) -> mismatch (an otherwise-trusted key with no issuer authorization is rejected, not allowed);
  - `issuer not in allowed_issuers` -> mismatch.
  This directly prevents the reported case: a key trusted for issuer A signing a manifest that claims issuer B. Since the mismatch is appended and `signature_verified` is never set, the result is `MISMATCH` with `signature_verified=False`.
- Opt-in semantics are reasonable: an empty `trusted_key_issuers` returns `None` (no enforcement), preserving backward compatibility. Worth being explicit in the spec/README that issuer binding is only enforced when the map is populated — with it empty, a valid key for any issuer still passes. That's the documented intent, just flagging for operator awareness; not blocking.
- The map is threaded through `VerifyRequest` -> `VerificationContext` -> `POST /verify` consistently.
- Tests are thorough: authorized (VALID), cross-issuer (A-key vs B-manifest -> MISMATCH), key-with-no-authorization -> MISMATCH, missing-manifest-issuer -> MISMATCH, plus the HTTP-level `test_http_post_verify_enforces_trusted_key_issuers`.
- Full CI matrix, CodeQL, lint, mypy, security scan all green.

PR #221, opened 2026-07-06
fix(verify): bind hybrid signatures to combined key id

### LINKED ISSUE #216: security: derive hybrid component keys from the registered combined key_id
Tracked follow-up from the June 2026 org-wide security hardening review. Wave 1/2 fixes are merged and published; this is remaining hardening.

In hybrid mode the verifier honors per-component key ids (ed25519_key_id / ml_dsa65_key_id) taken from the unsigned signature block, letting an attacker steer which trusted key each component is checked against. Derive the two component keys only from the registered combined key_id and verify sha256(ed_pub||pq_pub)==key_id; do not honor attacker-supplied per-component key ids.

carloshvp:
## Summary
- derive hybrid verifier component keys from the trusted combined public key registered under `signature.key_id`
- verify `sha256(ed25519_pub || ml_dsa65_pub) == signature.key_id` before checking hybrid signatures
- remove trust in unsigned per-component key ids and add no-`pyoqs` regression coverage for the key-selection path

## Why
Hybrid verification previously allowed the signature block to steer component key lookup via `ed25519_key_id` / `ml_dsa65_key_id`. Those fields are not the trust anchor. The verifier should bind both components to the registered combined key id and reject any trusted key material that does not hash back to that id.

Closes #216.

## Validation
- `PYTHONPATH=src python3 -m pytest tests/test_verify.py -q`
- `python -m pytest -q`
- `pytest -v --tb=short --cov=agent_manifest --cov-report=term-missing --cov-fail-under=80`
- `ruff check src/ tests/ --select E,F,W --ignore E501`
- `mypy src/agent_manifest`
- `bandit -r src/agent_manifest -c pyproject.toml`
- `pip-audit`

### DIFF
--- a/python/src/agent_manifest/_verify.py
+++ b/python/src/agent_manifest/_verify.py
@@ -189,6 +189,30 @@ class VerificationContext(BaseModel):
 
 # Manifest spec versions this verifier implementation can process (spec 2.4).
 SUPPORTED_MANIFEST_VERSIONS: frozenset[str] = frozenset({"0.1"})
+_HYBRID_ED25519_PUBLIC_KEY_BYTES = 32
+
+
+def _split_hybrid_public_key(
+    key_id: str,
+    public_key_bytes: bytes,
+) -> tuple[bytes, bytes]:
+    """Split and authenticate a combined Ed25519 || ML-DSA-65 public key."""
+    if len(public_key_bytes) <= _HYBRID_ED25519_PUBLIC_KEY_BYTES:
+        raise ValueError(
+            "Hybrid public key must be Ed25519 public key bytes followed by "
+            "ML-DSA-65 public key bytes"
+        )
+
+    actual_key_id = hashlib.sha256(public_key_bytes).hexdigest()
+    if not hmac.compare_digest(actual_key_id, key_id):
+        raise ValueError(
+            "Hybrid public key bytes do not match signature.key_id"
+        )
+
+    return (
+        public_key_bytes[:_HYBRID_ED25519_PUBLIC_KEY_BYTES],
+        public_key_bytes[_HYBRID_ED25519_PUBLIC_KEY_BYTES:],
+    )
 
 
 def _signature_key_issuer_mismatch(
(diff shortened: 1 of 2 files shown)


imran-siddique (dismissed):
Approving. This closes a real hybrid mix-and-match / key-substitution vector.

What I verified:
- `_split_hybrid_public_key` (`python/src/agent_manifest/_verify.py`) now derives BOTH component keys from the single combined public key registered under `signature.key_id`, and enforces `sha256(ed25519_pub || ml_dsa65_pub) == key_id` with `hmac.compare_digest` (constant-time) before any signature check. The prior code trusted attacker-controlled `ed25519_key_id` / `ml_dsa65_key_id` from the *unsigned* signature block to steer per-component key lookup — that let each component be pointed at a different trusted key (classic mix-and-match / downgrade). Those fields are now explicitly ignored. Good.
- No path where a component verifies against the wrong key: both `ed_bytes` and `pq_bytes` are non-overlapping slices of the same authenticated blob, and `HybridVerifier` (`_signing.py:416`) requires BOTH the Ed25519 and ML-DSA-65 components to pass over the same `signing_pre_image` — so neither component can be dropped or substituted to force a downgrade.
- Fail-closed: a registered key whose bytes don't hash back to `key_id` raises `ValueError`, caught into a `signature` MismatchDetail -> `OverallResult.MISMATCH`, `signature_verified=False`.
- Tests cover both the happy path (`test_hybrid_signature_uses_combined_trusted_key`, asserting the exact 32-byte Ed25519 / remaining ML-DSA split is what reaches the verifier) and the negative path (`test_hybrid_trusted_key_must_match_combined_key_id`). The negative test also exercises the key-selection path without pyoqs, since `_split_hybrid_public_key` rejects before `HybridVerifier` construction touches oqs.
- Full CI matrix (3.11/3.12/3.13 x ubuntu/macos/windows), CodeQL, lint, mypy, security scan all green.

Nice tight fix.

imran-siddique:
Thanks for this, it's approved with green CI. It's conflicting with main now (the branch has fallen behind). Could you rebase onto the latest main and force-push? Once it's mergeable it'll merge and auto-close #216. Happy to help if anything's unclear.

carloshvp:
@imran-siddique Rebased onto the latest main and force-pushed in 3fd4af5. I resolved the _verify.py conflict by preserving the trusted-key issuer authorization from current main while keeping hybrid signatures bound to the combined key ID. All CI checks are green, and the focused verification suite passes 43 tests.

The force-push cleared the previous approval, so the PR now needs re-approval when you have a moment.

imran-siddique (approved):
Approving. This closes the hybrid key-steering flaw from #216 cleanly:

- Component keys now derive solely from the combined key registered under `signature.key_id`, with `sha256(ed_pub || ml_dsa65_pub) == key_id` enforced via constant-time compare before the split. Attacker-supplied `ed25519_key_id`/`ml_dsa65_key_id` are correctly ignored.
- Negative test covers the trusted-key-that-doesn't-hash-back case (→ MISMATCH, no signature_verified). Happy path asserts the exact 32-byte split.
- Tight scope, all CI green.

Good fix, thanks @carloshvp.

Issue #272, opened 2026-08-04
spec_change: a VALID verification result is not authorization for the action that triggered it (sections 5.3, 7.2)


Mayur021:
The spec opens by saying an Agent Manifest proves "who the agent was, what it was allowed to do, how it was built, what it decided, who approved it, and whether any of that changed between approval and execution."

What the data model actually binds under "what it was allowed to do" is the tool catalog. `tool_manifest.catalog_hash` is a Merkle root over per-tool schema and description leaves, so it pins which tools were approved and it detects mutation of their definitions. Inside an authorized tool, a read and an irreversible write are the same tool. The manifest establishes which capabilities were approved. Nothing in it bounds what a call through one of them may do.

That is a defensible scope. My problem is that the scope is nowhere stated, and one sentence in 5.3 makes the opposite reading easy:

> A relying party receiving a `MISMATCH` MUST NOT proceed with the operation that triggered verification.

There is no matching sentence for VALID. A gateway reading that rule literally has defined behaviour on mismatch and no defined behaviour on match, and the natural default is to proceed. 5.3.1 already puts verification at the moment a runtime session binds. #267 puts a signed state snapshot at the Tool Use stop and says it supports credential-gated access decisions, which is the same boundary. At that boundary, establishing where the state came from and deciding whether the action may run are two different questions asked in the same millisecond, and only the first one has an answer here.

The spec handles the same distinction correctly in two other places, which is why I read this as an omission rather than a position:

- `delegation_chain` returns UNVERIFIABLE rather than VALID when a `scope_grant.constraints` element carries Cedar the verifier cannot evaluate. An unevaluated constraint does not quietly become a pass.
- `hitl_record` returns APPROVAL_INSUFFICIENT when an approval does not meet the `approval_method` bar for the declared `risk_tier`.

`risk_tier` is the closest thing the spec has to a consequence vocabulary, and it lives inside `approved_scope` on an approval record. It describes the approval given at issuance, not the action being attempted at runtime. `escalation_policy.trigger` is the other hook, and it is OPTIONAL free-form Cedar with nothing connecting it to that vocabulary.

What I am asking for, additive only:

1. One more bullet in 7.2. Close to: the manifest does not bound the consequence of an action a verified agent takes. It attests the approved capability set. Whether a particular call within that set may run, and what oversight that call needs, is an authorization decision outside this specification.

[...]

PR #280, opened 2026-08-10
Conformance test: what VALID does and does not establish

Mayur021:
What
Adds a test module asserting what `verify_manifest` establishes today, and what it does not.

Why
Per the discussion on #272. A manifest reaches VALID with every artifact intact and
`tool_manifest.catalog_hash` matching byte for byte, while the call made through an authorized
tool is one no reviewer would have approved. That is not a defect in verification. It is the
boundary of what verification answers: VALID establishes that the agent is the one that was
approved, and says nothing about what that agent may do with the tools it was approved to hold.

The structural form of the argument is that `verify_manifest(manifest, context,
revocation_store)` has no parameter that could carry an invocation, so the result cannot depend
on one even in principle. A SELECT and an unqualified DELETE through the same approved tool
produce identical results.

`test_every_verified_field_is_an_artifact_not_an_action` names the ten fields the result reports
today. If a field describing an action is ever added, that test fails and whoever adds it has to
decide what VALID now means.

Scope is deliberately narrow as requested: the read versus irreversible-write distinction inside
a single authorized tool. Parameter-level policy in general is a design question and is not
demonstrated here.

These tests pass against main as written, and they are intended to.

Spec impact
None. SDK tests only.

## Test plan

- [x] `pytest -v` passes — 6 new tests pass; full suite 634 passed, 37 skipped
- [ ] `mypy src/agent_manifest` passes — 5 pre-existing errors in
      _revocation.py on a clean main, none from this PR (tests only)
- [ ] `ruff check src/ tests/` passes — the repo has no ruff config, so this
      runs under defaults. This file trips 1 (datetime.now(timezone.utc));
      tests/test_am_verify.py trips 5 of the same. Left consistent with the
      module it sits beside rather than diverging.
- [x] Spec change: none, SDK tests only

## DCO

### DIFF
--- a/python/tests/test_valid_authorization_scope.py
+++ b/python/tests/test_valid_authorization_scope.py
@@ -1,0 +1,159 @@
+"""AM-VERIFY-SCOPE: what VALID does and does not establish - issue #272.
+
+These tests assert what the implementation does today, not what it should do.
+They pass against main as written, and they are intended to.
+
+A manifest can reach VALID with every artifact intact and its tool catalogue
+matching byte for byte, while the call made through an authorized tool is one
+no reviewer would have approved. That is not a defect in verification. It is
+the boundary of what verification answers: VALID establishes that the agent is
+the one that was approved, and says nothing about what that agent may do with
+the tools it was approved to hold.
+
+Scope is deliberately narrow, per the discussion on #272: the read versus
+irreversible-write distinction inside a single authorized tool. Parameter-level
+policy in general is a design question and is not demonstrated here.
+"""
+import inspect
+from datetime import datetime, timedelta, timezone
+
+import pytest
+
+from agent_manifest._signing import Ed25519Signer, generate_ed25519
+from agent_manifest._verify import (
+    FieldResult,
+    OverallResult,
+    RevocationStore,
+    VerificationContext,
+    verify_manifest,
+)
+
+NOW = datetime.now(timezone.utc)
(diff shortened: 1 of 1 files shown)


imran-siddique (approved):
Exactly the scope we agreed on #272, and better made than I expected.

The structural test is the one that carries it. Inspecting the signature to show there is nowhere to put a call proves the result cannot depend on one even in principle, which is a stronger claim than any pair of examples could make. The SELECT and the unqualified DELETE then make it concrete for a reader who wants to see it rather than reason about it.

`test_every_verified_field_is_an_artifact_not_an_action` is the part that will still be earning its place in a year. Enumerating the ten fields and asserting none describes an action means anyone adding an action-shaped field has to stop and decide what VALID now means, rather than quietly widening it.

Ran it against the local source: 6 passed.

One triviality, nothing to change: `tool_catalog_hash` is set as an attribute after the context is constructed rather than passed to the constructor. It is a real optional field so it behaves identically.

The normative half of #272 is still mine to decide and this does not pre-empt it. It makes the gap concrete for whoever reviews that, which is more useful before the decision than after.

Mayur021:
Thanks for the read. Leaving tool_catalog_hash as it is.

Agreed the normative half is yours and this does not touch it. Happy to look at whatever you land on if a second pair of eyes is useful, but I will not push it in the meantime.

Issue #256, opened 2026-08-01
Spec gap: no way to author a composition-only manifest where the model is unknowable


imran-siddique:
## Where this came from

Building an agent-integrity check for GitHub Copilot in [agentrust-io/integrations#68](https://github.com/agentrust-io/integrations/pull/68). It measures what a **repository** contributes to an agent's composition: `.github/copilot-instructions.md`, `.github/instructions/**`, `AGENTS.md` anywhere in the tree, `.github/skills/*`, `.claude/skills/*`, `.agents/skills/*`, and MCP config. Those files decide how the coding agent behaves, they arrive by pull request, and they are reviewable.

It cannot emit an Agent Manifest, and I think the reason is a genuine gap rather than a limitation of the integration.

## The problem

Level 0 requires **all artifact bindings** (section on levels: "Software-only | All artifact bindings"), and `artifacts.model_identity` is artifact 4.

A repository cannot know the model. Copilot's model is chosen at session time by the user's plan and settings. The same repository serves every model a user might select, and the composition the repository contributes is identical across all of them. `model_hash` is unavailable, and so is `provider` / `model_id` / `version` in any truthful form.

The obvious workarounds are all dishonest in a way this project usually refuses:

- `provider: github, model_id: copilot` describes a **product**, not a model.
- `model_id: unknown` asserts a binding to a thing named "unknown".
- Omitting the artifact makes the manifest non-conformant at every level.

So the integration currently emits nothing and claims no `integrates_with`, which I would rather fix in the spec than paper over in the integration.

## The part that suggests the spec already half-agrees

The verification result vocabulary (section 5.2) is:

```
"model_identity": "MATCH | PROVIDER_ASSERTED | MISMATCH | NOT_BOUND  -- REQUIRED",
```

`NOT_BOUND` exists for **every** artifact in that result object. But there is no legal way to author a manifest that produces it, because every level requires all bindings present. A verifier can report a state no conformant manifest can reach.

That reads like an internal inconsistency rather than a missing feature.

## What I would propose

Three options, roughly in order of how much I like them.

**1. A composition-only profile.** A manifest that binds a stated subset of artifacts and declares which ones it deliberately does not bind, with verification returning `NOT_BOUND` for those. Something like:

```json
{
  "@type": "AgentManifest",
  "profile": "composition-only",
  "unbound_artifacts": ["model_identity", "decision_trace", "memory_baseline"],
[...]

PR #302, opened 2026-08-14
test(cose): publish the COSE negative conformance vectors (#243 phase 3)

zohebk8s:
## Summary

Phase 3 of the ADR-0011 migration, tracked in #243. Publishes the COSE negative conformance vectors, `AM-VEC-COSE-002` through `008`, alongside the positive encoding vector that landed with phase 2.

The vectors are the portable contract, so these are what gate any other-language SDK.

## What is covered

| Vector | Expected | Rule under test |
|---|---|---|
| `AM-VEC-COSE-002` | `MISMATCH` | A tampered protected header invalidates the signature |
| `AM-VEC-COSE-003` | `MISMATCH` | `alg` in the unprotected header is rejected, never read |
| `AM-VEC-COSE-004` | `MISMATCH` | A vendor-tree `typ` alias is refused |
| `AM-VEC-COSE-005` | **`VALID`** | An unprotected header injected after signing does not change the verdict |
| `AM-VEC-COSE-006` | `MISMATCH` | An untagged structure is rejected rather than inferred |
| `AM-VEC-COSE-007` | `MISMATCH` | Trailing bytes after the COSE object |
| `AM-VEC-COSE-008` | `MISMATCH` | A detached payload; this profile is inline only |

The first four are the cases named in section 9 of the envelope specification. The last three cover places a CBOR implementation genuinely differs, so they catch cross-language divergence rather than restating v0.1 behaviour that the existing vectors already pin.

### Negatives carry no pinned encoding

The positive vector pins every element, because an implementation has to produce those exact bytes. A negative cannot: its bytes are malformed by construction, so pinning their decomposition would assert that a verifier can parse something it is being told to reject. Negatives therefore carry `envelope_hex` and an expected result, and no `expected.cose` block.

**The consequence is that the schema records that a manifest is rejected, not why.** Every structural rejection maps to `MISMATCH`, so a verifier that rejects one of these for the wrong reason still passes.

Two things mitigate that without changing the schema. Each vector names the rule under test in its `description` and `spec_refs`. And `test_cose_negative_vector_is_not_silently_unparseable` asserts that every mutated envelope still decodes as a four-element CBOR array, so a vector cannot quietly degrade into "rejected because it was garbage" while still appearing to test its rule.

Where a vector can isolate its rule, it does. `AM-VEC-COSE-004` is signed **over** the aliased header rather than having `typ` edited into an already-signed one, so its signature is valid and the `typ` value is the only defect. Built the other way, an implementation that verified the signature and never implemented the `typ` rule would have passed it. Each negative was checked individually against the reference verifier to confirm it is refused for the reason it names rather than incidentally:

```
AM-VEC-COSE-002  InvalidSignature
AM-VEC-COSE-003  alg present in the unprotected header
AM-VEC-COSE-004  typ is 'application/vnd.agent-manifest+cose', expected ...
AM-VEC-COSE-006  untagged COSE structure: expected CBOR tag 18
AM-VEC-COSE-007  trailing bytes after the COSE object
AM-VEC-COSE-008  payload must be a byte string, inline not detached
```

An optional `expected.reason` would close the gap properly, but that is an addition to a published contract, so I would rather raise it here than assume it. Happy to add it if you want the negatives to assert cause as well as outcome.

### One vector expects `VALID`

`AM-VEC-COSE-005` injects an unprotected header after signing and expects `VALID`. That is the same rule as `003` seen from the other side: nothing in the unprotected header is covered by the signature, and section 6 step 7 evaluates it last, so an injection must not change the verdict. A verifier that merged the two halves, or read `kid` from the malleable one, fails it.

It sits with the negatives because it tests the same normative sentence, but it is worth knowing before someone reads a `VALID` in a file named for a negative case.

### No post-quantum vector

ML-DSA-65 signing is hedged. `cryptography` exposes `sign(data, context=None)` with no deterministic mode, and signing identical bytes twice produces different signatures, so a post-quantum envelope cannot be pinned the way `AM-VEC-COSE-001` is.

A vector whose bytes changed on every regeneration would be a snapshot rather than a contract, and shipping a private seed for consumers to regenerate against would break the suite's own rule that no private key material is written to disk. The README records this so the omission is a decision rather than something rediscovered later.

## Not included

Regenerating the suite also updates all twenty v0.1 vectors, because `intent` joined `SIGNED_FIELDS` in 0.11.0 after those vectors were last generated. I confirmed that drift is present on `main` independently of this change, and left it out to keep this diff scoped. Happy to send it as its own PR.

## Consuming a negative vector

`tests/vectors/README.md` documents the shape. It is the same as a positive vector, decode `envelope_hex` and run the verifier over the bytes; only the assertion differs. The two properties above, that negatives state outcome rather than cause and that one of them expects `VALID`, are called out there as well, since the README is what another language reads.

## Verification

| Check | Result |
|---|---|
| Full suite | 874 passed, 6 skipped |
| Conformance suite | 37 passed |
| mypy strict | clean |
| bandit | clean |
| ruff, `--select E,F,W --ignore E501` as CI runs it | all checks passed |
| Vectors reproducible from the fixed seed | byte-identical on regeneration |

Refs #243


imran-siddique (changes requested):
The seven new envelope-shape vectors are well constructed, but this cannot yet serve as the phase-3 portable contract described by the PR and issue. The phase-2 security follow-up on #243 explicitly assigns five additional negative vectors to phase 3: (1) unauthorized issuer/signing-key pairing, (2) duplicate JSON member names, (3) non-finite JSON numbers, (4) manifest-version/envelope mismatch, and (5) excessive payload nesting. None is present in the generator or index.

These are not generic v0.1 repetitions: each covers a COSE-path defect found during phase 2 that another-language implementation following the earlier envelope document could reproduce. In particular, outcome conformance without the issuer/key vector does not enforce the authorization boundary, and ordinary JSON parsers commonly diverge on duplicates/non-finite values.

Please add deterministic COSE vectors for all five cases, list them in the index/README, and include them in the regeneration and reference-verifier tests. Each vector should isolate its named defect where feasible (valid signature over the malformed/unauthorized payload rather than incidental signature failure). CI is green for the current set, but the claimed cross-language security contract is incomplete without the cases already recorded in #243.

zohebk8s:
All five added, plus two more the findings on #243 imply. Each is in the generator, the index and the README, and each is signed over the payload under test so the rule it names is the only reason to reject it.

| Vector | Rule | Expected |
|--------|------|----------|
| `AM-VEC-COSE-009` | Trusted key not authorized for the manifest's issuer | `MISMATCH` |
| `AM-VEC-COSE-010` | Duplicate JSON member name | `MISMATCH` |
| `AM-VEC-COSE-011` | Non-finite literal `NaN` | `MISMATCH` |
| `AM-VEC-COSE-012` | `0.1` payload in a `0.2` envelope | `INCOMPATIBLE_VERSION` |
| `AM-VEC-COSE-013` | Payload nested past the depth bound | `MISMATCH` |
| `AM-VEC-COSE-014` | `0.2` manifest falling back to the v0.1 envelope | `MISMATCH` |
| `AM-VEC-COSE-015` | Non-finite literal `Infinity` | `MISMATCH` |

**009** is byte-identical to `AM-VEC-COSE-001`; only `context.trusted_key_issuers` differs. Nothing about the object explains the rejection, so a verifier without an authorization boundary returns `VALID`.

**014 and 015 are not in your list**, but both are in the findings.

`014`: your follow-up records the version gate as bidirectional, and `012` was only one direction. It expects `MISMATCH`, which is what #274 shipped and `python-v0.11.0` released: the verifier does support 0.2, it just will not verify 0.2 through the v0.1 path, so an unsupported-version result would state something untrue about its capabilities. The vector publishes that behaviour rather than proposing a change to it.

`015`: finding 2 records `NaN` and `Infinity` as one class of defect. They are not one code path in every parser, so a verifier that special-cases `NaN` passes `011` and still accepts `Infinity`. `-Infinity` travels the same path as `Infinity`, so it is covered rather than given a third vector.

**Isolation is asserted, not claimed.** Every negative carries `signature_valid`, re-derived from the bytes on disk using only `keys.json`. Separately, stripping the named defect from each of the seven must produce `VALID`, which is what rules out a second defect. Both checks select on vector id, so a vector that forgets to declare either one fails rather than dropping out.

**One case cannot be a vector.** The hybrid half of finding 1, one authorized component key alongside one unauthorized one, needs an ML-DSA-65 signature. That signing is hedged and `cryptography` 49 has no deterministic mode, so the bytes differ on every regeneration, and exempting it from the regeneration check would undermine the rest of the suite. It stays covered by `test_every_hybrid_signer_must_be_authorized_for_the_issuer`, and the README states the rule as binding on other languages.

**Two fixes found along the way.** A new regeneration test caught pre-existing drift on `main`: since `intent` joined `SIGNED_FIELDS` in 0.11.0, nineteen v0.1 vectors published a `signed_fields` list without it. No signature or result changes, but it is the field other languages build their pre-image from. Separately, `index.json` declared `spec_version` 0.1 while fourteen vectors target the v0.2 envelope.

**One thing to decide separately.** `010`, `011`, `013` and `015` impose rules section 6 does not state; step 3 only says to parse the payload as JSON. Right now the vectors are the sole normative source. Happy to raise a section 6 issue.

35 vectors, 935 passed and 6 skipped, green across Python 3.11 to 3.13 on Linux, macOS and Windows. Merged `main` to clear a CHANGELOG conflict; the vectors still regenerate byte-identically after it.

zohebk8s:
Merged `main` to clear a CHANGELOG conflict. That added 20 regenerated vector files to the diff, so a note on why.

`SIGNED_FIELDS` gained `profile` and `unbound_artifacts` in #306. Every signed manifest records a copy of that list in `signature.signed_fields`, and the committed vectors publish it, so the vectors and the SDK no longer agreed. The regeneration test in this branch caught it on the merge, the same way it surfaced the earlier `intent` addition from #288.

Nothing verifies differently, which is why this is easy to miss. `signing_pre_image` skips fields absent from the manifest rather than serializing them as null, so a manifest declaring neither field produces byte-identical signed bytes and every existing signature still verifies. No signature, envelope bytes or expected result moves here. The change is one list per file:

```diff
   "crypto_profile",
+  "profile",
+  "unbound_artifacts",
   "artifacts",
```

What was stale is the list another language would read to build its pre-image, which is the part the vectors exist to publish.

944 passed and 6 skipped, vectors regenerating byte-identically.

zohebk8s:
Merged `main` again, same CHANGELOG conflict, same regeneration.

`SIGNED_FIELDS` gained `source_bundle` in #307. That is the third addition in this sequence without a matching regeneration: `intent` in 0.11.0, `profile` and `unbound_artifacts` in #306, `source_bundle` now. Each time the committed vectors kept publishing a `signature.signed_fields` list that no longer matched the SDK, and each time the regeneration test in this branch caught it on the merge.

Nothing verifies differently in any of the three. `signing_pre_image` skips fields absent from the manifest rather than serializing them as null, so a manifest declaring none of them produces byte-identical signed bytes and every existing signature still verifies. No signature, envelope bytes or expected result has moved. What goes stale is the list another language would read to build its pre-image, which is the part these vectors exist to publish.

Three times in three weeks looks like a property of the workflow rather than of any one change: adding to `SIGNED_FIELDS` breaks nothing, fails no test, and leaves the vectors quietly disagreeing with the SDK. Once this lands, the test runs on `main` and the next one fails in its own PR instead of being found here.

958 passed and 6 skipped, vectors regenerating byte-identically.

imran-siddique (approved):
The requested phase-3 coverage is now present: unauthorized issuer/key binding, duplicate JSON members, non-finite numbers (NaN and Infinity), version/envelope mismatch, and excessive nesting. Focused vector suite: 86 passed; exact CI lint rule passed. The full local suite is shadowed by an unrelated installed #256 checkout, so hosted CI remains the complete-suite authority. Approving.

PR #308, opened 2026-08-18
spec+sdk: prove the audit chain grew by appending (#273)

### LINKED ISSUE #273: Decision-trace binding (§3.2.7) has no post-signing continuity mechanism, unlike memory (§3.2.6) and the COSE post-signing artifacts
**Component:** agent-manifest spec, v0.2 draft (`spec/agent-manifest-spec-v0.2.md`) and the v0.2 COSE envelope (`spec/agent-manifest-cose-envelope-v0.2.md`, ADR-0011, 2026-07-27), on `main`.

**Type:** spec change proposal — a gap where a claim scoped to post-signing decisions rests on a binding fixed at signing time, while the same document solves the equivalent problem for two other artifacts.

**Scope (do not drop):** this is a spec read, not an implementation report. If a later revision already carries a continuity mechanism for the audit chain and I missed it, this does not apply.

**Process note:** CONTRIBUTING lists the spec-change file as `spec/agent-manifest-spec-v0.1.md`, but the sections below are from the v0.2 draft and the v0.2 COSE envelope on `main`. Flagging in case section numbering differs in the canonical file; the gap is about the mechanism, not the numbering.

## Summary

The abstract states a verifier holding a manifest and its attestation report can "prove, without trusting the operator, that a specific agent instance ran specific code under specific policy with specific tools, **produced specific decisions**, and received specific human oversight."

The "produced specific decisions" claim is carried by the decision-trace binding (§3.2.7). That binding fixes `audit_chain_root` "as of manifest signing time." Decisions are produced after signing. Nothing in the binding lets a third party confirm that the audit chain served at `audit_chain_uri` at a later time is an append-only extension of the root signed at T0 — without trusting the operator who serves that endpoint. The claim is scoped to post-signing behaviour; the mechanism is scoped to signing time.

## Why this is a gap and not a design choice

The same specification solves exactly this problem for the two other artifacts that change or attach after signing:

- **Memory (§3.2.6).** A memory advance is accepted as governed only if an RFC 9162 §2.1.2 consistency proof shows the prior `memory_root` is an append-only positional prefix of the new one. Explicitly: "absence of a valid consistency proof never grants acceptance." Fail-closed. A verifier can confirm post-signing growth without trusting the operator.
- **Receipts, attestation report, HITL approvals (COSE envelope §4).** These attach after signing in the COSE unprotected header, each with its own integrity story — receipt signed by the Transparency Service, attestation by the TEE, approval by its approver.

The decision trace is the one artifact that grows after signing and receives neither treatment. It is bound as a snapshot root at T0 (§3.2.7) plus a URI the operator serves. `entry_count` (OPTIONAL) "allows verifiers to detect entry deletion" — it detects removal below the signed count, but nothing binds the shape of growth above the signed root.

## Precise gap

§3.2.7 provides, as of signing time: `audit_chain_root`, `first_entry_at`, `last_entry_at`, `entry_count`. A verifier fetching the chain at T+n can check it is internally consistent and that its head still contains the signed root, but cannot verify that the chain at T+n is an append-only extension of the T0 root rather than a rebuilt or selectively rewritten chain that happens to re-contain the signed prefix. That verification requires a consistency proof between the signed root and the current head. §3.2.6 requires exactly that for memory; §3.2.7 does not require it for the audit chain.

## Suggested fix

Apply the §3.2.6 pattern to §3.2.7: define a checkpoint on the audit chain and require an RFC 9162 §2.1.2 consistency proof for any advance beyond the signed `audit_chain_root`, so a verifier can confirm the chain at T+n descends from the root signed at T0 without trusting the operator. This reuses a mechanism the spec already specifies, tests with a vector, and relies on for memory — it is not new machinery.

Alternatively, scope the abstract's claim to what a signing-time root supports ("the audit chain as of deployment") and treat the growing decision record as a separately-committed artifact with its own continuity mechanism, the way the COSE envelope already treats receipts and approvals.

imran-siddique:
Closes #273. Part of the CoSAI WS4 Phase 1 review (ws4 #149) close-out.

## The gap

`decision_trace.audit_chain_root` (§3.2.7) is the chain state at manifest signing time. Decisions happen after that. A verifier fetching the chain at T+n could check it was internally consistent and still contained the signed root, and could not distinguish an append-only extension from a chain rebuilt around that prefix. `entry_count` detects deletion *below* the signed count and constrains nothing above it.

@contrario filed this against the v0.2 draft, and the second half of the report is what makes it a spec bug rather than a wish: the same document already solves this problem for memory (§3.2.6.2, RFC 9162 consistency proof, fail-closed) and for the artifacts that attach after signing in the COSE envelope. The audit chain was the one artifact that grows after signing and got neither.

## What this does

- **spec §3.2.7.1** — audit checkpoint over `{audit_chain_root, tree_size, seq, observed_at, ttl_seconds}`, continuity evidence per `trace_type`, four ordered acceptance rules, published test vectors for both trace types.
  - `merkle-log`: RFC 9162 §2.1.2 consistency proof, O(log n).
  - `hash-chained`: the ordered entry leaves appended after the signed position, folded forward from the signed root, O(n). A sequential chain admits no logarithmic proof; a linear proof is the honest alternative to no proof.
- **`_audit_continuity.py`** — leaf construction, both fold shapes, and `verify_continuity` with verdicts `accepted | discontinuity | rollback | expired`.
- **`decision_trace` gains `EXTENDED`** in §5.2 / `FieldResult`. Roots identical is `MATCH`; a proven append-only extension is `EXTENDED`; absent, malformed, or failing evidence stays `MISMATCH`. This adds a proven path, it does not soften the unproven one.
- **`AuditCheckpointBinding`** in `models.py`, mirroring `MemoryCheckpointBinding`.

## Three decisions worth arguing with

1. **No budget stage.** A memory delta has one. An audit chain is meant to grow without bound, so a growth ceiling would be a constraint the artifact cannot honestly carry. Symmetry with §3.2.6.2 was not worth a fake rule.
2. **`0x02` for the sequential fold, not the RFC 9162 internal-node `0x01`.** With `0x01`, a left-spine merkle tree and a hash chain over the same entries produce identical roots for n ≤ 3. The verifier is pinned to the signed `trace_type`, so this is not an attack, but a root that can only be read one way is worth one byte. `test_the_two_trace_types_never_share_a_root` holds it.
3. **`entry_count` is now load-bearing.** It is the only chain size the issuer signed, so a runtime anchoring its proof at a different signed size is choosing its own baseline. Rejected before the proof is examined.

## What it does not claim

A verified proof establishes that nothing below the current head was rewritten or removed. It does not establish that every action produced an entry. Completeness against real behaviour stays out of scope (§7.2) and §3.2.7.1 says a continuity proof MUST NOT be presented as evidence of it. That is the distinction the Phase 1 review record settled on, and it would have been easy to lose here.

## Verification

32 new tests: published vectors for both trace types, the rebuilt-chain case `entry_count` cannot catch, truncation, algorithm switch, malformed roots, seq rollback in both shapes, TTL expiry, and the `EXTENDED` path end to end through `verify_manifest`. Full suite 1103 passed, 6 skipped. `mypy` strict and `ruff` under the CI rule set both clean.

@imran-siddique for spec-editor and maintainer review — this touches `/spec/` and the verifier, and I have not self-approved.


Qiang-Xu (approved):
Looks good to me. Just like `agt` so far the merkle tree is in memory, would we persist it in the future?

Issue #265, opened 2026-08-03
[spec] Artifact-only refresh invalidates manifest_hash_in_report without re-attestation


sapsan14:
## Spec section

Sections 2.2 (Manifest Lifecycle / Artifact-only refresh), 3.3
(`manifest_hash_in_report` pre-image), and 3.2.6.2 (memory checkpoint/delta
binding).

## Problem

Section 2.2 currently permits this exception:

> Artifact-only refresh: Used only for `memory_baseline.snapshot_hash` renewal
> within the same `ttl_seconds` window. The manifest is re-signed but the TEE
> attestation need not be re-run if no other artifact has changed.

That exception is incompatible with the normative hash construction in
Section 3.3. `manifest_hash_in_report` is computed over the RFC 8785
serialization of the full manifest including the `signature` block and signed
artifact bindings, excluding only `attestation` and the later
`transparency_log_entry`.

Changing `memory_baseline.snapshot_hash` and re-signing therefore changes the
pre-image twice: the artifact binding changes and so does the signature. The
old TEE attestation still binds the old hash. A conforming verifier must either
reject the refreshed manifest or stop enforcing the hash binding. Both cannot
be valid at once.

This is especially confusing now that v0.2 defines a separate append-only
memory checkpoint/delta protocol in Section 3.2.6.2. That protocol is the
natural place for governed memory evolution; it should not silently mutate the
root document whose exact hash was attested.

## Proposed change

1. Remove the artifact-only refresh exception for the root manifest at Level
   1 and above. Any change to a signed field, including
   `memory_baseline.snapshot_hash`, MUST produce a new manifest and a fresh
   attestation binding its new `manifest_hash_in_report`.
2. Permit evolving memory without root re-attestation only through a distinct
   checkpoint/delta evidence object under Section 3.2.6.2. Define its signer,
   link to the attested base manifest, expiry/freshness rule and verification
   result separately; do not call the updated checkpoint a refreshed root
   manifest.
3. Add a normative negative vector: change the snapshot hash, re-sign, retain
   the old `attestation` block, and require verification to fail.
4. Clarify whether Level 0 software-only manifests may use the legacy refresh
   shortcut, since they have no hardware hash binding to invalidate.

## Impact

- Backward compatible: no for implementations relying on the stated Level 1+
  shortcut; yes for implementations already using fresh attestation or the
  checkpoint/delta protocol
- Affects conformance level(s): Level 1, Level 2 and Level 3
[...]

PR #310, opened 2026-08-18
spec+sdk: agent_id is the stable one, instance scope is declared (#269)

### LINKED ISSUE #269: spec_change: designate agent_id as the correlation key for OCSF ai_agent.instance_uid (§3.1, §5.2)

imran-siddique:
Closes #269. Part of the CoSAI WS4 Phase 1 review (ws4 #149) close-out, and the item the Phase 1 review record listed as the gate on any normative correlation rule.

## What was blocking

@rabbidave proposed that OCSF `ai_agent.instance_uid` MUST equal the manifest `agent_id`. #285 landed the crosswalk informative instead, because `agent_id` was serving both of OCSF's identity roles — the stable `uid` and the session-scoped `instance_uid` — and a MUST would have forced a stable identifier into the field OCSF explicitly defines as the non-stable one. §6.4.4 recorded the open question rather than guessing at it.

## The resolution

`agent_id` is the **stable** identity, full stop. A deployment may still encode an instance in the SPIFFE path, but a verifier MUST NOT infer scope from it — §3.1 calls that structure a convention, and a convention is not something a consumer can read meaning out of.

Instance scope is **declared**: a new OPTIONAL, signed `agent_instance_id` (UUID v7). Its absence means the manifest governs every run of `agent_id`.

§6.4.2 becomes normative. A producer emitting `ai_agent` for a manifest-governed session:

1. MUST populate `uid` from `agent_id`
2. MUST populate `instance_uid` from `agent_instance_id` when the manifest declares one
3. MUST NOT fall back to `agent_id` for `instance_uid` when it does not
4. MUST carry `manifest_id` where the event class can express it

## Why this is not the mapping the issue asked for

Rule 3 is the one that matters. Binding `instance_uid` to `agent_id` makes the two OCSF fields identical for every deployment following the stated convention, and a consumer cannot separate "every run of this agent" from "this run" or even tell the collapse happened. Emitting no `instance_uid` is a better failure than emitting a stable one.

The issue was right that a join key was missing and that `instance_uid` is where it belongs. Splitting the declaration is how that key exists without one field carrying two lifetimes.

## Signed, not decorative

`agent_instance_id` joins `SIGNED_FIELDS`. An instance identity outside the signature is one an operator can retarget after the fact, which breaks exactly the correlation an evidence consumer depends on. Same compatibility argument as `intent`: the pre-image omits absent fields, so every existing signature verifies unchanged — and there are tests for both halves, that an absent field leaves the pre-image byte-identical and that a declared one changes it.

Conformance vectors are regenerated for the longer `signed_fields` list; `AM-VEC-018`'s `manifest_hash_in_report` moves with it. That is the drift the regeneration test from #302 exists to catch.

## Also

§5.2 gains a `correlation` object (`agent_uid`, `agent_instance_uid`, `manifest_id`), populated from the document rather than earned by passing verification — an auditor investigating a `MISMATCH` needs the join key most.

§6.4.3 delegation stays informative and says why: populating `delegation.uid` needs a per-hop durable identifier §3.4 does not have, plus an answer on issued versus self-asserted authority. That is a data-model change and is not proposed here.

## Verification

11 new tests. Full suite 1082 passed, 6 skipped. `mypy` strict and `ruff` under the CI rule set both clean. Vectors byte-identical on regeneration.

@imran-siddique for spec-editor and maintainer review — normative text plus a `SIGNED_FIELDS` change, and I have not self-approved. @rabbidave, this rejects the literal proposal and keeps its substance; the argument is in §6.4.4 if you want to push back on it.


Mayur021:
The split is the right call, and rule 3 is the part I'd defend hardest. Binding instance_uid to agent_id would have made the two OCSF fields identical for every deployment following the convention, and a consumer could not even tell the collapse had happened. Declaring scope beats parsing it out of a path segment that carries no declared meaning.

Two things before this goes normative.

Signing agent_instance_id runs into the existing re-issuance rule. Section 2 says any change to a signed field MUST produce a new manifest, and at Level 1 and above a fresh attestation binding a new manifest_hash_in_report. So an instance-scoped manifest costs a new manifest_id, a TEE attestation run and a transparency log entry for every run it governs, and under #265 the one path that skipped re-attestation was withdrawn at Level 1 and above, so there is no lighter route for a signed-field change. I am not arguing the field should be unsigned, since your retargeting point is correct. I think the spec should say what the intended operating model is: instance-scoped manifests for long-lived or high-risk runs, with multi-run manifests staying the norm for conversational volume? Whatever the answer, an implementer will hit this on the first deployment and it is better answered in 6.4.2 than in an issue later.

Second, 6.4.2 gives three answers to what a producer emits for instance_uid when the manifest declares no agent_instance_id. Rule 3 says only what it must not be. The prose says emitting nothing is the better failure. The crosswalk table says the producer's own session identifier. Those are not the same instruction, and the table is the one most likely to be implemented. Worth picking one and saying whether the producer's own session identifier is SHOULD or MAY.

Smaller: 5.2 marks correlation.agent_instance_uid REQUIRED while the SDK models correlation itself as Optional and only populates it when agent_id is a non-empty string. Since agent_id is REQUIRED in the schema, is there a case where a v0.2 verification result legitimately has no correlation object? If not, a consumer can rely on it always being there and 5.2 could say so.

Also flagging that it currently shows as conflicting against main.

pforest (approved):

imran-siddique:
The current spec now states the operating model in §3.1: instance-scoped manifests for long-lived or high-risk runs, with multi-run manifests the normal model for conversational volume. In §6.4.2, both the table and rule 3 now say SHOULD use the producer’s session identifier and MAY omit it when unavailable; neither permits agent_id as the fallback.

PR #314, opened 2026-08-18
spec+sdk: bind a verification result to the request that asked for it (#266)

### LINKED ISSUE #266: [spec] Signed verification results need relying-party challenge and context binding
## Spec section

Sections 5.1 (Verification Endpoint), 5.2 (Verification Result Schema), 5.3
(Verification Semantics), and 3.3.2 (Runtime Attestation Reports / Freshness
Proofs).

## Problem

The remote verification protocol has no relying-party challenge. The request
does not carry a nonce, and the signed result does not bind the requester,
`purpose`, `required_fields`, `enforce_hitl`, `enforce_attestation`, or the
other verification-context choices that determined `VALID`.

`verification_id` is generated by the verification service and `verified_at`
is a producer-selected timestamp. Neither proves to a relying party that the
result was created for its live request. A previously signed `VALID` result can
therefore be replayed:

- to a different relying party;
- for a different purpose, such as substituting an `audit` result for a
  `tool-call` decision;
- after the relying party requested a stronger set of fields or enforcement
  flags; or
- after runtime state changed but before a local age heuristic expires.

Section 3.3.2 correctly introduces a verifier nonce for the separate runtime
attestation report, but Sections 5.1–5.3 do not carry or bind that challenge
into the main signed verification result. The two freshness domains are thus
not composed.

## Proposed change

1. Add a `challenge_nonce` supplied by the relying party to the verification
   request (minimum 128 bits of cryptographic randomness, single-use).
2. Add the nonce and a normative
   `verification_context_hash = SHA-256(JCS(verification_context))` to the
   verification result. The result signature MUST cover both.
3. Echo or otherwise bind the relying-party `verifier_id`, `purpose`,
   `required_fields`, enforcement flags and any policy/version reference that
   affects the decision.
4. Require relying parties to verify that the nonce equals their outstanding
   challenge, has not been consumed, and the response is within a locally
   bounded acceptance window. A producer timestamp alone MUST NOT satisfy the
   freshness check.
5. When `enforce_attestation=true`, bind the Section 3.3.2 runtime report to
   the same challenge nonce, or to a domain-separated nonce derived from it,
   and include the runtime-report digest in the signed verification result.
6. Add negative conformance vectors for cross-verifier replay,
   cross-`purpose` substitution and weaker-`required_fields` substitution.

## Impact

- Backward compatible: likely requires a protocol-version increment for
  portable signed results; local in-process verification can retain an
  explicitly non-portable profile
- Affects conformance level(s): all levels that expose or consume a remote
  signed verification result
- Conformance tests that need updating: AM-VERIFY and AM-ATTEST; add replay and
  context-substitution negative vectors
- Regulatory mapping impact: none claimed; this narrows the meaning of a
  verification result to the relying party's actual decision context

## Alternatives considered

- Rely only on `verified_at` plus a short TTL: rejected because the timestamp
  is producer-controlled and does not bind the result to a specific request or
  relying party.
- Rely only on TLS: rejected because verification results and evidence packs
  are designed to be signed, stored and passed across trust boundaries; TLS
  does not prevent later replay or context substitution.
- Use a service-generated request ID: insufficient unless the relying party
  supplies unpredictable challenge material and checks single use.

## AI assistance disclosure

OpenAI Codex assisted with repository inspection and first-draft wording. The
finding was checked against spec v0.2 at commit
`f0c54ac7d40f70fb5c8d170cd7e89eb3f3728c4a` and the current implementation and
test suite.

imran-siddique:
Closes #266. Part of the CoSAI WS4 Phase 1 review (ws4 #149) close-out. Normative text carried by a maintainer per `GOVERNANCE.md#who-may-author-normative-text`; @sapsan14 reported it.

## The gap

`verification_id` is generated by the verification service; `verified_at` is a producer-selected timestamp. Neither shows a relying party that a result was produced for **its** live request. So a signed `VALID` is replayable — to a different relying party, for a different `purpose`, against a weaker context than the one now being asked for, or after runtime state changed but before a local age heuristic expires.

§3.3.2 already defines a verifier-supplied nonce for the separate Runtime Attestation Report. The two freshness domains were never composed.

## What lands

New **§5.1.2**, six rules:

1. A freshness-dependent request supplies `challenge_nonce` (≥128 bits, CSPRNG, single-use); the service echoes it unchanged.
2. The result carries `verification_context_hash` over the RFC 8785 canonical form of the request context, and **every** input that affected the decision must be inside it.
3. The nonce is echoed as its own field and is *not* an input to that hash.
4. `verification_signature` covers both — a signature over a result whose nonce sits outside it authenticates the verdict, not the question.
5. A relying party that supplied a challenge checks the nonce is its outstanding one, unconsumed, inside a local window, and recomputes the context hash. A producer timestamp alone MUST NOT satisfy freshness.
6. When `enforce_attestation` is set and a runtime report is used in the same decision, its nonce is `sha256("am-runtime-nonce" || challenge_nonce)` — exposed as `derive_runtime_nonce()`.

## Two decisions worth arguing with

**The nonce is deliberately outside the context hash.** The relying party compares it directly; keeping it out means two requests asking the same question hash the same, which is the only thing that makes the value comparable across requests.

**`challenge_nonce` is optional, not required.** No existing caller breaks, and a caller reading a historical manifest is not forced to invent freshness it does not need. What changes is that a result with no nonce now *says so by omission*, and §5.3 states what it may be relied on for: an audit reading, not a gate on an action. Silence with a documented meaning beats a required field callers would fill with a constant.

## What the SDK hashes, and what it does not

`_CONTEXT_HASH_FIELDS` lists the inputs this implementation lets affect a decision. Runtime artifact hashes and `trusted_keys` are excluded on purpose: they are the evidence being checked and the anchors doing the checking, not the question the relying party asked, and folding them in would make the hash change for reasons the relying party cannot reproduce from its own request. One test per included field, so a context field added later and forgotten there fails as missing coverage.

**Not implemented:** `required_fields` is in the spec's §5.1 request context and this SDK does not act on it, so binding it here would assert a decision input that does not exist. Rule 2 requires an implementation that supports it to include it.

## Verification

24 new tests. Full suite 1095 passed, 6 skipped. `mypy` strict and `ruff` under the CI rule set clean.

@imran-siddique for spec-editor and maintainer review — normative protocol text, and I have not self-approved.


imran-siddique:
@lywinged — all required checks are green. Could you take the maintainer approval pass when you have a moment?

lywinged (commented):
Took the pass. Two things I'd want closed before this lands, and one wording point on rule 2 itself. Not approving yet — auto-merge is armed here, so an approval from me merges it rather than queueing a review, and these should reach you first.

**1. Rule 5 cannot be satisfied against the SDK as written.**

Rule 2 hashes the `verification_context` object *as received*. Rule 5 has the relying party recompute that digest from the context it sent. `verification_context_hash()` hashes a locally reconstructed subset of `VerificationContext` instead, and the two field sets differ in both directions:

```
in the request, never hashed : required_fields
hashed, never in the request : conformance_level, require_delegation, strict_artifact_verification
```

Sending exactly the §5.1 body, and recomputing per rule 2 with a stock JCS library:

```json
"verification_context": {
  "purpose": "tool-call",
  "verifier_id": "spiffe://trust.example/auditor",
  "required_fields": ["system_prompt", "policy_bundle"],
  "enforce_hitl": true,
  "enforce_attestation": false,
  "min_slsa_level": 3
}
```

```
relying party : sha256:20f73759f9000b331509387c8dfa45bb7b6e0a086e7c06b0c1845c3d1ced9016
service       : sha256:d0eec3e8db48ae79b0d4c8abd52af258cc9f8a723ed0244f4008236df8e261c0
```

Rule 5 then obliges that party to reject. It fails closed, so this is interop and conformance rather than a downgrade — but a check that can never pass is a check that gets switched off, and rule 5 goes quietly with it.

The last three are verifier-side policy. They do affect the decision, and they are still the wrong thing to hash: they are not the question the relying party asked, and it cannot reproduce a digest over settings it never sent. That is the same criterion this PR already uses to exclude `trusted_keys` and the runtime hashes.

`test_a_relying_party_can_recompute_the_hash_from_what_it_sent` does not catch it because it hands the relying party the verifier's own `VerificationContext` object, so it asserts `f(x) == f(x)`. A real relying party holds the wire request and nothing else.

**2. Rule 6 is a MUST with no enforcement path.**

`derive_runtime_nonce` has no caller inside the package, and `verify_runtime_report(report, nonce, context_hash)` accepts whatever nonce it is given. A report bound to an independently chosen nonce passes today — the composition failure rule 6 exists to forbid. Adding a keyword-only `challenge_nonce=` that requires the nonce to be the rule 6 derivation makes the correct path the easy one without changing any existing call.

**3. Rule 2's wording, while this is still a draft.**

Two readings, and they produce implementations that cannot verify each other:

- *"every input that affected the decision MUST be inside that object, **including** `purpose`, `verifier_id`, `required_fields`, …"* — is that list unconditional, or illustrative? This SDK omits `required_fields` on the second reading; a conformance suite has to pick one.
- The sentence also merges what a relying party can **recompute** with what it can only be **told**. Rule 5 only works for the first. They cannot be the same field.

Separately: this SDK's canonicalizer applies NFC normalization per §4.3, and RFC 8785 does not normalize at all. So "the RFC 8785 canonical JSON of the `verification_context` object" is not literally what this implementation computes, and a relying party using a stock JCS library diverges on any string that is not already NFC — a `verifier_id` with an internationalized path reaches it.

I have a patch for 1 and 2, with the relying-party test the suite is missing. Full suite unchanged against baseline. Happy to open it, or take the description and write it your way.

The rest reads right to me. The gap is real — `verification_id` and `verified_at` are both producer-chosen and neither is evidence of anything. Rule 3's reasoning is not obvious and is correct. And making the nonce optional while 5.3 states what an unbound result may be relied on for is the honest shape.

pforest (dismissed):

Qiang-Xu:
Right, in the spec I see `required_fields` but I don't see it in the implementation here.

lywinged:
> Right, in the spec I see `required_fields` but I don't see it in the implementation here.

Yes, and it is deliberate rather than an oversight, the PR description says so under "What the SDK hashes, and what it does not": this SDK does not act on `required_fields`, so hashing it would assert a decision input that does not exist.

That is a defensible reading, and it is what point 3 of my review was about. Rule 2 says "every input that affected the decision MUST be inside that object, **including** `purpose`, `verifier_id`, `required_fields`, …". Read the list as unconditional and the SDK is short a field; read it as illustrative and it is not. A conformance suite has to pick one, and the spec currently does not say which.

The half with no answer yet is the other direction. The hash also covers `conformance_level`, `require_delegation` and `strict_artifact_verification`, none of which are in the §5.1 request context — so a relying party never sent them and cannot reproduce a digest over them. Rule 5 has it recompute and reject on mismatch, which means it rejects every result, `required_fields` or not.

Separately, and bigger than either: #322.

Issue #298, opened 2026-08-13
Gap report: governed memory checkpoint advances have no behavioral regression assessment


nikshilov:
## Spec section

- Section 3.2.6 - Memory Baseline Binding
- Section 3.2.6.2 - Memory Checkpoint and Delta Protocol
- Related discussion: #254

## Problem

The current v0.2 draft, section 3.2.6.2, can establish that a candidate checkpoint is an append-only extension of an approved memory state, that its sequence advances, that it is fresh, and that it stays within a declared update budget. These are necessary integrity and governance properties. They do not establish that retrieval behavior remained acceptable.

The repository's `docs/memory-governance.md` states this boundary directly: memory validity is defined through integrity, drift, and freshness rather than semantic usefulness.

A governed checkpoint can therefore pass the current checks while:

- a correction is appended, but the superseded fact remains more salient;
- authorized additions displace a required safety, identity, or continuity anchor;
- memory from another role, project, task, or tenant becomes retrievable;
- states that should produce different context collapse to the same retrieval.

No signature forgery, checkpoint rewrite, expired TTL, or invalid consistency proof is required for these failures. This is also narrower than evaluating whether a final model answer is correct. The property under assessment is the context selected from a candidate memory checkpoint before that checkpoint is approved.

## Proposed change

Start with non-normative tooling and evidence rather than a new required Agent Manifest field.

The proposed first release is an external `MemoryCheckpointAssessment/0.1` artifact plus a reference harness and deterministic test vectors. The artifact would bind:

- approved baseline and candidate checkpoint digests;
- probe-suite digest;
- retriever identity and version;
- deterministic metrics and assessment result;
- evaluation time and signer provenance.

The harness would compare the baseline and candidate through a small retriever adapter and test explicit invariants for correction precedence, anchor preservation, scope isolation, and state-conditioned differentiation. A model judge may help with exploratory diagnosis, but it would not be the sole source of the deterministic assessment result.

Public fixtures would include the complete data, probes, and pinned retriever needed for reproduction. For private stores, the appraisal would state the limit clearly: checkpoint digests provide identity and provenance, but an independent rerun also requires access to the same snapshot contents, probe suite, and retriever.

An initial Agent Manifest example could carry the appraisal beside checkpoint evidence without changing current validity rules. Whether a future specification hook is useful would remain a maintainer and design-partner decision.

Public implementation work I would build from:

- https://github.com/zbs-gg/pulse-paper
[...]

PR #316, opened 2026-08-18
spec: say what a governed memory checkpoint does not establish (#298)

imran-siddique:
Refs #298 (stays open — see below). Normative-adjacent scope text carried by a maintainer; @nikshilov filed the report.

## The gap

§3.2.6.2 establishes that a candidate checkpoint is an append-only extension of an approved state, that its sequence advances, that it is fresh, and that it is within budget. It establishes nothing about retrieval behaviour over the new state, and nothing in the protocol looks at that.

Every failure mode in the report passes all four checks as written:

- an appended correction stays less salient than the fact it corrects;
- authorized additions displace a safety, identity, or continuity anchor;
- memory from another role, project, or tenant becomes retrievable;
- states that should retrieve differently collapse to the same context.

None needs a forged signature, a rewritten checkpoint, an expired TTL, or an invalid consistency proof. The protocol's own checks are structurally unable to see any of them.

`docs/memory-governance.md` already drew that line; the spec did not. A reader of §3.2.6.2 alone could reasonably take "governed advance" to mean the memory is still fit to retrieve from.

## What this does not do

It does not add the `MemoryCheckpointAssessment/0.1` artifact the report proposes. That is an external non-normative deliverable needing a reference harness, a pinned retriever adapter, and deterministic public fixtures, and the report is right that it starts outside the manifest rather than as a required field.

**#298 stays open** as the tracking issue for that work. Writing the boundary down first is what stops the spec claiming an assessment before the harness exists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,18 @@
   defined value space and no verification behaviour, which is what it always
   was; the field name resembled an assurance signal and nothing said otherwise.
 
+- **[SPEC]** Stated the scope boundary of the memory checkpoint protocol
+  (section 3.2.6.2, issue #298). A governed advance establishes integrity,
+  ordering, freshness and budget, and establishes nothing about retrieval
+  behaviour over the new state. A checkpoint can satisfy every rule while an
+  appended correction stays less salient than the fact it corrects, while
+  authorized additions displace a safety or identity anchor, while another
+  tenant's memory becomes retrievable, or while states that should retrieve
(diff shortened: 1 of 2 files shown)


imran-siddique:
@zohebk8s — all required checks are green. Could you take the maintainer approval pass when you have a moment?

zohebk8s (approved):

zohebk8s (approved):

PR #317, opened 2026-08-18
spec+sdk: bind a behavioural assessment to the system prompt (#254)

### LINKED ISSUE #254: Spec change proposal: bind a configuration assurance result to the system prompt
## Spec section

- Section 3.2.1 (System Prompt Binding), primary
- Section 7.1 / 7.2 (Threat Model, Out of Scope)
- Section 3.2.5 and 3.2.5.1 (RAG Corpus Binding), cited as existing precedent
- Section 5.2 / 5.3 (Verification Result Schema and Semantics)

## Problem

Section 3.2.1 binds the system prompt by digest:

```json
"system_prompt": {
  "hash": "sha256:<64-hex-chars>",
  "hash_algorithm": "SHA-256 | SHAKE-256",
  "version": "...",
  "classification": "public | internal | confidential | restricted",
  "language": "...  -- OPTIONAL",
  "safety_level": "<operator-defined safety tier>  -- OPTIONAL",
  "bound_at": "..."
}
```

Every field is metadata or integrity. The one safety-adjacent field, `safety_level`, is operator-defined, OPTIONAL, appears exactly once in the specification, and carries no definition, no constraint on its values, no normative rule, and no verification behaviour. It is an operator assertion rather than evidence.

The threat model treats this artifact as covered. T1 reads:

> T1 - Prompt Substitution | System prompt replaced in memory between approval and runtime
> | Undetectable. No binding exists. | `system_prompt.hash` mismatch immediately detectable
> at verification.

T1 to T10 are each a substitution, mutation, or forgery performed by an adversary. The prompt is *replaced*, the policy *swapped*, the tool definition *mutates*, documents are *injected*, a record is *fabricated*. Section 7.2 excludes five further threats, the nearest being "semantic prompt injection at the model layer", which again presumes adversarial input.

There is a hazard in this artifact that is neither addressed in 7.1 nor excluded in 7.2: **a system prompt authored in good faith, reviewed, approved, signed, sealed to a TEE measurement, and verified VALID at every conformance level, which specifies behaviour the deploying organisation would reject if it were tested rather than read.**

No adversary is present. No digest changes. Every control in the specification passes. By the spec's own description, artifact #1 defines "persona, behavioral boundaries, safety constraints". It is not data the agent consumes, it is the specification of the agent's behaviour, and its hazard is therefore expressible in the approved value itself. For this artifact, digest equality is necessary but not sufficient, and no verification result distinguishes *approved and assessed* from *approved and unassessed*.

The specification already recognises this class of problem for exactly one artifact. Section 3.2.5 binds the RAG corpus by `merkle_root` and additionally carries `poisoning_scan` with `scanner_version`, `scanned_at`, `result`, under normative force: a manifest with `result: flagged` MUST NOT be issued as VALID, and `not-scanned` is not permitted at Level 1 and above. That is an acknowledgement that for some artifacts content can be hazardous without substitution, and that the remedy is a bound assessment rather than a stronger hash.

Across all ten artifacts, exactly one carries a measurement. Two carry an adjective: `safety_level` on artifact #1 and `safety_alignment_version` on artifact #4.

## Demonstration

I built a Level 2 manifest binding a real vendor system prompt, 321 bytes, digest `sha256:1aef8354770be432157ada310a0c15945dbe982d35b6ed28d34d251c11783eaf`. It is ordinary commercial text for a sales assistant. Nothing in it is malicious and no attacker is involved anywhere in what follows.

The behaviour of that exact artifact was measured against the applicable policy across four frontier models (Claude Opus 5, Gemini 3.5 Flash, Kimi K3, GLM 5.2), 25 scenarios spanning five role and data-class combinations and five pressure levels, both with the prompt bound and against a no-prompt baseline. 200 responses, 186 scored by two independent judges with 98.4% agreement:

| | violation rate |
|---|---|
| configuration bound | 62% (n=88) |
| neutral baseline | 4% (n=98) |
| **delta** | **+58 points** |

At pressure level zero, with an ordinary non-adversarial request, the delta is +39 points. All four models show the effect.

Verified against upstream v0.7.0 at conformance level 2:

```
A. manifest as issued today (no assurance_test)
   result            : VALID
   signature_verified: True
   system_prompt     : MATCH
```

A manifest carrying an `assurance_test` block is rejected outright with `extra_forbidden`, so the remaining cases cannot be expressed by this SDK at all.

Against a build carrying the proposed change, same manifest, same digest, same level:

```
A. manifest as issued today            result: MISMATCH   system_prompt: MATCH
B. assessment ran and flagged          result: MISMATCH   system_prompt: MATCH
C. assessment passed with provenance   result: VALID
```

`system_prompt: MATCH` holds throughout. Integrity passes in every case, correctly. The manifest is exactly what was approved. The question the specification cannot currently ask is whether what was approved behaves acceptably.

Scenario set, policy definition, and judge specification are from the CAID benchmark (CAID Benchmark v1.2, Andrey Lazarev, 2026, MIT), used unmodified so the violation definition is the benchmark's rather than mine.

## Proposed change

Extend Section 3.2.1 with an `assurance_test` block, modelled on `poisoning_scan`:

```json
"assurance_test": {
  "harness": "<harness name and semver, format: tool-name/x.y.z>  -- REQUIRED for Level 2+",
  "tested_at": "<ISO 8601 UTC>  -- REQUIRED for Level 2+",
  "baseline": "<identifier of the comparison condition tested against>  -- REQUIRED for Level 2+",
  "result": "pass | flagged | not-assessed  -- REQUIRED"
}
```

Normative rules, mirroring 3.2.5.1:

- `result: flagged` MUST NOT be issued as VALID. The manifest MUST be held INCOMPLETE until the configuration is revised and a passing assessment completed. (Wording mirrors 3.2.5.1. Note that the SDK's existing poisoning path returns MISMATCH rather than INCOMPLETE for `flagged`; the reference implementation mirrors that behaviour rather than diverging from it, so if the normative wording and the implementation should be reconciled, that applies to both artifacts and is worth settling here.)
- At Level 2 and above, `not-assessed` is NOT permitted, and an absent block is treated as `not-assessed`.
- Below Level 2, `not-assessed` is permitted but MUST surface as a warning. An absent block below Level 2 MUST verify silently, so manifests issued before this change are unaffected.
- At Level 2 and above, `result: pass` MUST carry `harness`, `tested_at`, and `baseline`.
- `harness` MUST include tool name and semantic version as `<tool-name>/<semver>`.

`baseline` is the load-bearing field. An assessment of a configuration is only meaningful against a declared comparison condition, because the quantity of interest is the behavioural delta the configuration introduces rather than an absolute score. Without it a result is self-reported and not reproducible by a third party.

Additionally, a verification status `ASSURANCE_NOT_ASSESSED` for `system_prompt` when the artifact is correctly bound and attested but carries no assessment. This follows `PROVIDER_ASSERTED`, which exists so a weaker form of evidence surfaces to the verifier instead of silently returning VALID.

Harness identity is left open, consistent with `poisoning_scan.scanner_version` and with the Section 10.2 roadmap item for an AAIF-registered scanner identifier format. The same registry would serve both fields.

## Impact

Verified rather than asserted. Reference implementation on a branch off `941ff4e`:

https://github.com/zohebk8s/agent-manifest/tree/spec/assurance-test-artifact-1

295 insertions, zero deletions, DCO signed off. CI green on all ten jobs: the test matrix across Python 3.11, 3.12 and 3.13 on Ubuntu, macOS and Windows, plus ruff, mypy, bandit, pip-audit, AGT governance verify, and the build check. A runnable two-verifier demonstration and a full test report are available if useful.

I have not opened a pull request. CONTRIBUTING.md puts conformance level changes in the issue-and-discussion track, so the branch exists to be inspected rather than merged, and I would rather agree the shape here first.

- **Backward compatible:** yes, in the direction that matters. A pre-change manifest re-validates unchanged, silently, below Level 2.
- **Not forward compatible.** Because every spec object sets `extra="forbid"`, a manifest carrying `assurance_test` is rejected by a pre-change verifier with `extra_forbidden`. This needs a coordinated version bump and cannot ship as an opt-in vendor extension.
- **Affects conformance level(s):** Level 2 and Level 3 normative, Levels 0 and 1 warning only.
- **Conformance tests that need updating:** none existing. 19 new tests added in the style of `test_poisoning_scan.py`. Full suite went from 626 passed / 18 skipped to 645 passed / 18 skipped, no regressions.
- **Implementation surface:** 60 lines of source (16 in `models.py`, 44 in `_verify.py`), plus 235 lines of tests, for the 295 total above. `VerificationContext.conformance_level` already exists to gate the 3.2.5.1 rules, and `_verify` already contains the level-aware block that raises a `MismatchDetail` on `result: flagged`. This reuses that machinery for a second artifact.
- **Regulatory mapping impact:** EU AI Act Art. 15 and Art. 9 currently map to integrity controls only. A bound pre-deployment assessment is closer to what those articles ask for than a digest is.

One implementation note worth flagging. My first cut emitted a warning whenever the block was absent, which broke three existing tests in `test_poisoning_scan.py` asserting `warnings == []` at Levels 0 and 1. An absent block and an explicit `not-assessed` are different states: the first means the manifest predates the change, the second means the deployer declared no assessment. Treating them identically breaks backward compatibility. The rules above reflect the corrected behaviour.

## Note on why this must be a spec change

`models.py` documents the intent directly: unknown fields are rejected because the spec defines exhaustive field sets per object, so an unrecognised key is structural drift rather than an extension point. I confirmed this against 0.7.0. `assurance_test` is rejected, as are the conventional extension conventions (`x-` prefix, underscore prefix, a generic `ext` object).

That is a reasonable design decision and I am not asking for it to change. It does mean there is no way to carry this evidence, or to pilot the idea in a deployment, without a normative change to Section 3.2.1.

## Scope

Artifact #1 only.

Artifacts #3 (tool descriptions) and #6 (memory baseline) share the relevant property, that their content enters the model's context and shapes output, and would be candidates for the same treatment later. Artifact #5 already has it.

Deliberately out of scope: artifact #2, whose correctness is a static analysis problem with a different remedy, already noted in `LIMITATIONS.md`; and artifact #4, which is model evaluation and belongs to a different discipline.

## Alternatives considered

**Constrain `safety_level` instead.** Simpler, but it stays an operator assertion. The distinction here is between a claim about a configuration and a reproducible result from a named harness against a declared baseline, a line the spec already draws for the corpus.

**Carry this as a TRACE claim instead.** Rejected. TRACE v0.2 section 2.4 makes model behaviour a permanent scope boundary, correctly: TRACE adjudicates what executed, not whether output was correct. This proposal is deliberately not runtime output adjudication. It is a pre-deployment property of a configuration artifact, fixed before signing, in the same category as a corpus poisoning scan.

**Defer to AGT's PromptDefense Evaluator.** PromptDefense is a 12-vector injection audit and `agt red-team scan` is adversarial probing. Both measure resistance to a hostile party. Neither assesses whether the deployer's own approved configuration specifies acceptable behaviour, and neither result binds into the manifest.

**Treat this as out of scope and document it in 7.2.** A legitimate outcome, and better than the present silence. But given that 3.2.5 already solves this problem shape for a lower-risk artifact, excluding it for artifact #1 is hard to explain to an auditor asking why the knowledge base is scanned and the behavioural specification is not.

imran-siddique:
Closes #254. Normative text carried by a maintainer per `GOVERNANCE.md#who-may-author-normative-text`; @zohebk8s reported it, with measurements rather than an argument, which is why this is a mechanism change and not a documentation one.

## The evidence in the issue

A real vendor system prompt, 321 bytes, ordinary commercial text, no attacker anywhere: **62% violation rate** with the configuration bound against **4%** on a neutral baseline. 25 scenarios, four frontier models, 186 responses, two independent judges at 98.4% agreement, +39 points of delta even at pressure level zero. The manifest verifies `VALID` at Level 2 throughout — correctly, because the manifest *is* exactly what was approved.

## The gap in the threat model

T1 through T10 are each a substitution, mutation, or forgery by an adversary, and `system_prompt.hash` detects those. The nearest exclusion in §7.2 is semantic prompt injection, which also presumes adversarial input. Nothing addresses or excludes a prompt authored in good faith, approved, signed, sealed and verified that specifies behaviour the deploying organisation would reject if it were tested rather than read.

Artifact #1 is not data the agent consumes; it is the specification of the agent's behaviour. Its hazard is expressible in the approved value itself, so digest equality is necessary and not sufficient.

## The spec already accepts this shape

§3.2.5 binds the RAG corpus by `merkle_root` **and** carries `poisoning_scan` under normative force: `result: flagged` MUST NOT be issued as `VALID`. New §3.2.1.1 is that pattern applied to artifact #1 — same rule, same reasoning.

## The half that matters more than the flagged rule

No verification result distinguished *approved and assessed* from *approved and unassessed*, so silence read as a pass. §5.2 gains `configuration_assurance` with `PASSED | FLAGGED | NOT_ASSESSED`, and `NOT_ASSESSED` is the default rather than a state a manifest opts into.

## Deliberately not decided

Whether Level 2 should require `result: passed`, as Level 1 already requires a scanned corpus, is a decision about what conformance costs every adopter. Not settled here, and the section says so. **That one is yours as sponsor** — everything else in this PR stands without it.

## `safety_level`

The issue is right that this field was the tell: operator-defined, OPTIONAL, one appearance in the whole spec, no value space, no constraint, no verification behaviour, and a name that reads as an assurance signal. It stays, because deployments use it as a routing label, and §3.2.1 now says plainly that a verifier MUST NOT treat it as evidence.

## What `passed` does not mean

It asserts that a named suite at a named version found nothing — not that nothing is there. `suite_id` and `suite_version` are REQUIRED for that reason: a result whose suite cannot be identified is not reproducible and carries no more weight than `safety_level` did.

## Verification

14 new tests. Full suite 1118 passed, 6 skipped. `mypy` strict and `ruff` under the CI rule set clean.


zohebk8s (approved):
Read the spec text, `models.py`, `_verify.py` and the tests. Looks right to me. Two calls
here are better than what I proposed: REQUIRED `suite_id` and `suite_version` put the
comparison somewhere inspectable rather than asserted, and the conditional `assessed_at`
closes a gap I left. `test_the_block_is_inside_the_signing_pre_image` is the one I would
most have wanted and did not think to write.

Three small things, none blocking:

**`evidence_digest` and `evidence_uri` are unenforced.** 3.2.1.1 says the digest MUST be of
the record at `evidence_uri`, but they are independent Optionals with no validator and
neither appears in the tests. A digest with no URI passes the schema with nothing for the
MUST to bind to. Everything else in the block is validated and tested, so this reads as an
oversight.

**3.2.5.1 and 5.2 now differ.** 3.2.5.1 holds a flagged `poisoning_scan` in `INCOMPLETE`;
5.2 says `FLAGGED` MUST cause `MISMATCH`, and the implementation agrees with 5.2.
Pre-existing, but sharper now that a second section states it and an implementer reading
both gets two answers.

**The absent-versus-`not-assessed` warning is not normative.** The reasoning in
`test_absent_block_produces_no_warning` is right, but 5.2 does not mention it, so a second
implementation would not reproduce it.

Leaving the conformance-level requirement open reads right.

PR #319, opened 2026-08-18
spec: replace raw NUL with escaped byte notation (#318)

### LINKED ISSUE #318: spec: a literal NUL byte makes the spec file un-greppable
spec/agent-manifest-spec-v0.2.md contains one raw NUL byte, at byte offset 52998, line 652. Verified through both the contents API and raw.githubusercontent, byte-identical, so it is in the file rather than an artefact of how I fetched it.

The line reads, with the byte shown as an escape:

Entry leaves. An entry leaf is H(0x00 || "am-trace-entry\x00" || canonical_entry), where canonical_entry is the RFC 8785 canonical JSON...

The first separator is written as text, 0x00. The second is an actual NUL inside the string literal.

Two consequences, and the first is why I am filing it rather than mentioning it.

grep treats a file containing a NUL as binary and prints nothing at all, not a zero count. So grep risk_tier spec/agent-manifest-spec-v0.2.md returns silence whether the term is present or absent. I hit this twice this afternoon: after #311 merged I searched for 5.3.2, got nothing, and concluded the section had not landed. It had. Only grep -a or reading the file in Python shows it. Anyone searching this spec for a field, a requirement or a section is getting an unreliable answer with no indication.

Second, the construction is ambiguous to a reader. This is a domain-separation definition, where one byte in the tag changes the hash and therefore whether verification succeeds. A reader cannot tell from the rendered line whether the tag is am-trace-entry or am-trace-entry with a trailing NUL.

The implementation settles the intent, so there is no design question here. python/src/agent_manifest/_audit_continuity.py line 49 has:

_ENTRY_TAG = b"am-trace-entry\x00"

so the trailing NUL is correct and the code writes it as the escape. The spec means the same thing and encoded it as the raw byte.

Fix: replace the raw byte with \x00 so the spec uses the same notation as the implementation it describes.

One smaller thing in the same file, which I raised on #311 before merge and which shipped as-is: 5.3.2 sits above 5.3.1 in document order, at lines 1340 and 1354. Heading level is correct, ##### matches 5.1.1 and 5.2.1; it is only the order. Mentioning it here rather than separately since both are one-line file fixes rather than content changes.

Happy to send a PR for either or both if that is easier than doing it yourself.

imran-siddique:
Closes #318.

## What changed

- Replaced the literal NUL byte in the trace-entry domain-separation tag with the textual `\x00` notation.
- Kept the specified bytes unchanged while making the Markdown file safe for ordinary text search.

## Root cause

The intended trailing NUL was embedded directly in the Markdown source instead of being represented as an escape, causing grep-compatible tools to classify the specification as binary.

## Validation

- Byte-level check: NUL count changed from 1 to 0.
- `rg risk_tier spec/agent-manifest-spec-v0.2.md` now returns normal text matches.
- `PYTHONPATH=python/src python -m pytest -q --basetemp C:\Users\imran\source\.pytest-tmp-am318`: 1118 passed, 6 skipped.
- `git diff --check`: clean.

### DIFF
--- a/spec/agent-manifest-spec-v0.2.md
+++ b/spec/agent-manifest-spec-v0.2.md
@@ -687,7 +687,7 @@ A checkpoint binds `{audit_chain_root, tree_size, seq, observed_at, ttl_seconds}
 }
 ```
 
-Entry leaves. An entry leaf is `H(0x00 || "am-trace-entry " || canonical_entry)`, where `canonical_entry` is the RFC 8785 canonical JSON (Section 4.3) of the audit entry. The tag domain-separates a trace entry from a memory operation or a corpus document that canonicalizes to the same bytes.
+Entry leaves. An entry leaf is `H(0x00 || "am-trace-entry\x00" || canonical_entry)`, where `canonical_entry` is the RFC 8785 canonical JSON (Section 4.3) of the audit entry. The tag domain-separates a trace entry from a memory operation or a corpus document that canonicalizes to the same bytes.
 
 Continuity evidence. The shape depends on the signed `trace_type`, and the verifier MUST use the `trace_type` bound in the manifest rather than one asserted at fetch time.
 


imran-siddique:
@Qiang-Xu — all required checks are green. Could you take the maintainer approval pass when you have a moment?

pforest (approved):

Qiang-Xu (approved):
Looks good to me!

PR #304, opened 2026-08-15
feat(tpm): add type-safe NV certify attestation parsing

noah-ing:
## What

Add a shared `TPMS_ATTEST` dispatch parser and a type-enforcing `TPM_ST_ATTEST_NV` entry point for `TPMS_NV_CERTIFY_INFO`.

## Why

This addresses the NV-certify consolidation scope added in #255. cMCP currently carries a local parser because `parse_tpm_quote()` assumes the quote union and rejects `TPM_ST_ATTEST_NV`. The new API lets consumers inspect the signed common header and safely parse a full NV certify without reimplementing TPM wire offsets.

The high-level `parse_tpm_nv_certify()` checks the signed attestation type before interpreting the union. The lower-level union parser remains available for callers that already dispatched on `TpmAttest.attest_type`. Size-prefixed `TPM2B_ATTEST` input must consume the full outer buffer; undeclared trailing bytes now fail closed.

This is the NV-certify portion of #255 only. It does not claim to close the remaining digest-agility or verification work.

## Spec impact

None. SDK-only; no normative or conformance-level change.

## Test plan

- [x] `pytest -q tests` - 874 passed, 6 skipped
- [x] `mypy src/agent_manifest` passes
- [x] `ruff check src tests --select E,F,W --ignore E501` passes
- [x] `bandit -r src/agent_manifest -c pyproject.toml` passes
- [x] New tests cover bare and TPM2B framing, a non-empty qualified signer, type confusion rejection, truncation, trailing bytes, public exports, and the maximum UINT16 offset
- [x] `CHANGELOG.md` updated

## DCO

All commits in this PR are signed off (`git commit -s`). By submitting this PR I certify the [Developer Certificate of Origin](https://developercertificate.org/).

## Author name correction

My full name is **Noah Ingwers** (`@noah-ing`). The original commit [`0d32ed8`](https://github.com/agentrust-io/agent-manifest/commit/0d32ed8fab639fa004ae691727e18a8393a84a93) incorrectly used `Noah Ing` in its author, committer, and `Signed-off-by` fields. Those entries refer to me, Noah Ingwers.

PR #320, opened 2026-08-18
feat(tpm): verify marshalled signature envelopes (#255)

### LINKED ISSUE #255: Absorb TPMT_SIGNATURE parsing so cmcp and ca2a can drop their copies
Follow-up from cmcp#447 (retiring the triplicated TPM verifier) and ca2a#73.

## The gap

`verify_tpm_quote` takes a **bare** signature. But the format real TPM tooling emits is `TPMT_SIGNATURE`: that is what `tpm2_quote -s` writes and what tpm2-pytss `signature.marshal()` returns. agent-manifest models `TPMS_ATTEST` (`parse_tpm_quote`) and the chain plus signature verification, but nothing unwraps the signature envelope.

So every consumer that collects a quote from actual hardware has to parse `TPMT_SIGNATURE` itself before it can call the canonical verifier:

- `cmcp_verify/tpm.py` has `parse_tpmt_signature` (added in cmcp#448)
- `ca2a_verify/tpm.py` now has its own (ca2a#74)

That is the same wire-format parser in two places, which is a smaller version of exactly the problem cmcp#447 set out to retire. Both were written against the same spec and both are load-bearing: without them the canonical verifier cannot be called on real evidence at all.

## Proposal

Export `parse_tpmt_signature(blob) -> ParsedSignature` (fields: `sig_alg`, `hash_alg`, `signature`) covering the three schemes a quote can use:

- `TPM_ALG_RSASSA` (0x0014) and `TPM_ALG_RSAPSS` (0x0016): size-prefixed `TPM2B_PUBLIC_KEY_RSA`
- `TPM_ALG_ECDSA` (0x0018): two size-prefixed integers, R then S, re-encoded as a DER sequence so `cryptography` verifies it directly

Optionally, accept a marshalled `TPMT_SIGNATURE` directly in `verify_tpm_quote` by sniffing the envelope, which would let callers pass hardware output through untouched.

Worth noting while this is open: `verify_tpm_quote` currently hardcodes SHA-256 for both the ECDSA and RSA paths, and RSA is PKCS#1 v1.5 only. Azure vTPM signs RSASSA/SHA-256 so that combination is validated, but a quote using RSAPSS or a SHA-384 digest would not verify. If `TPMT_SIGNATURE` parsing lands here, the `hash_alg` it carries is the natural way to select the digest instead of assuming one.

## Sequencing

Not urgent, and deliberately not blocking: ca2a#74 shipped with a local copy so the collector work was not gated on a release here. When this lands, both repos drop their copies and bump their floor.

/cc @imran-siddique

imran-siddique:
Closes #255.

## What changed

- `verify_tpm_quote()` now accepts marshalled `TPMT_SIGNATURE` envelopes and `ParsedSignature` values directly.
- Verification selects RSASSA, RSAPSS, or ECDSA from `sig_alg`.
- Verification selects SHA-256, SHA-384, or SHA-512 from `hash_alg`.
- Existing bare-signature callers retain SHA-256 compatibility.
- Signature envelopes with trailing bytes, unsupported hashes, or an algorithm that does not match the AK key type fail closed.

## Why

The parsers and NV-certify dispatch had already landed through #258 and #304, but the canonical verifier still ignored the envelope algorithm identifiers. Consumers therefore could not pass real TPM output through untouched and RSAPSS or SHA-384 quotes could not verify.

## Evidence

- Pre-change focused run: 5 new cases failed.
- Post-change focused run: 6 passed.
- TPM suite: 44 passed.
- Full suite: 1123 passed, 6 skipped.
- `mypy src/agent_manifest`: passed.
- CI Ruff command: passed.
- `git diff --check`: passed.

### DIFF
--- a/python/src/agent_manifest/_tpm_verify.py
+++ b/python/src/agent_manifest/_tpm_verify.py
@@ -12,8 +12,9 @@ Verification is fail-closed, in four steps:
 2. The attestation-key (AK) certificate chain is verified up to a
    caller-supplied trusted root (leaf issued-by next, root pinned by
    SHA-256 fingerprint). A self-signed AK is never trusted on its own.
-3. The AK signature over the ``TPMS_ATTEST`` blob is verified (ECDSA-P256 or
-   RSA PKCS#1 v1.5, both over SHA-256).
+3. The AK signature over the ``TPMS_ATTEST`` blob is verified. Bare signatures
+   retain the legacy SHA-256 behavior; a ``TPMT_SIGNATURE`` selects RSASSA,
+   RSAPSS, or ECDSA and SHA-256, SHA-384, or SHA-512 from its signed envelope.
 4. The qualifying data (the verifier's nonce, carried in ``extraData``) and the
    PCR digest (the platform measurement) are checked against expected values
    with constant-time compares.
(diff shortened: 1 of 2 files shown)


imran-siddique:
@carloshvp — all required checks are green. Could you take the maintainer approval pass when you have a moment?

pforest (approved):

PR #321, opened 2026-08-18
fix(hitl): define and enforce risk tiers (#313)

### LINKED ISSUE #313: spec_change: risk_tier sets the approval bar but is never defined, assigned or bound
risk_tier is load-bearing. It decides the required approval_method, it is what APPROVAL_INSUFFICIENT is measured against, it is a precondition of VALID when enforce_hitl is true, and it carries the Art. 14 mapping in section 7.

This is not #272 again. That one is about a call: VALID says nothing about the invocation that triggered verification. This one is about a deploy-time declaration inside the manifest that gates a requirement the spec itself imposes, and that no verifier checks. #272 is answered by putting authorization in the policy layer, which #311 does. This one is not, because the tier is consumed by the manifest layer rather than the policy layer.

The spec never defines it. low | medium | high | critical appears in the schema block and nowhere else as anything but a value to compare against. There is no statement of what the tiers mean, no rule for assigning one, and nothing a verifier can check the assignment against.

So it is a self-declared string that sets a security requirement. That is the one shape this document exists to eliminate. system_prompt, policy_bundle, tool_manifest, model_identity and the rest are bound to running artifacts and mismatch loudly. risk_tier binds to nothing and cannot mismatch, because there is nothing to compare it to.

The concrete failure: an operator declares low on an agent that can issue payments. A software-key approval is accepted, hitl_record does not return APPROVAL_INSUFFICIENT, verification returns VALID, and the Art. 14 evidence claim stands. No field mismatches, because the field that was wrong is the one nothing checks.

I am not proposing a new field, for the reason 5.3.2 gives: consequence is a property of a call, and anything carried in a deploy-time signed payload is stale by construction. What is missing is smaller. The spec should state what the tiers mean and on what axis they are assigned, so a declared tier is falsifiable by a reviewer even where it is not checkable by a verifier.

The axis that makes it falsifiable is reversibility rather than severity, because severity is a judgement and reversibility is a property of the action: whether the effect can be withdrawn, and whether withdrawal reaches outside the system that performed it. An action that cannot be undone, or whose effect has already reached a third party, does not become low risk because an operator felt it was routine.

Happy to draft the text if useful. Per GOVERNANCE.md#who-may-author-normative-text I am not the sponsor of it.

imran-siddique:
## Summary

- define risk tiers on a falsifiable reversibility and operational-boundary axis
- require the highest applicable tier and fail closed to critical when reversibility cannot be established
- state explicitly that VALID proves the signed declaration and method ordering, not that the tier is truthful
- make Level 2 verification return APPROVAL_INSUFFICIENT for software keys and for non-hardware high/critical approvals

Closes #313.

## Evidence

- focused HITL tests: 7 passed
- full Python suite: 1120 passed, 6 skipped
- regression cases separate the concrete failure (software-key/high) from the valid boundary (hardware-key/high)
- git diff --check: clean

## Environment note

Local mypy reaches the unchanged optional oqs import without a stub; repository CI is the authoritative configured type gate.

### DIFF
--- a/python/src/agent_manifest/_verify.py
+++ b/python/src/agent_manifest/_verify.py
@@ -911,6 +911,7 @@ def verify_manifest(
             # Check if any approval has expired (HITL-001: parse failure must set all_ok=False)
             now = datetime.now(timezone.utc)
             all_ok = True
+            approval_insufficient = False
             for approval in approvals:
                 approved_at = approval.get("approved_at", "")
                 duration = approval.get("approved_scope", {}).get("approval_duration_seconds", 0)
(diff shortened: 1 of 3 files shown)


imran-siddique:
@podcastinator — all required checks are green. Could you take the maintainer approval pass when you have a moment?

Mayur021:
The spec text answers what I filed. The axis is stated, the assignment rule is falsifiable by a reviewer without a verifier, highest-tier-wins closes the span case, and the paragraph separating what a verifier proves from what a reviewer checks keeps this consistent with 5.3.2. Defaulting to critical when the boundary cannot be established is the part I would have argued for and did not have to.

Two questions on the code rather than the text.

verify_manifest takes a raw dict and reads approvals with .get(), while models.py declares risk_tier as required and enum-typed on ApprovedScope. On the Level 2 path, is the schema validated before this check runs? If an approval can arrive without risk_tier, the value is None, the high-or-critical branch never fires, and mfa-backed returns APPROVED. That is the same shape as the issue itself, a rule keyed to a field that nothing on that path checks.

Second, the docstring says approvals are read from the unprotected header and are not covered by the payload signature. If that is the path a Level 2 verifier uses, what stops risk_tier being rewritten downward in transit?

pforest (approved):

PR #323, opened 2026-08-19
test(tpm): add reference-simulator signature vectors

devdiv07:
## What

Adds TPM quote/signature vectors generated by the Microsoft/TCG TPM 2.0
reference simulator and verifies them against Agent Manifest's canonical TPM
verification path.

This PR now contains **no production-code changes**. After rebasing onto #320,
the upstream verifier implementation is retained unchanged.

The corpus covers:

- RSASSA / SHA-256
- RSAPSS / SHA-256
- RSASSA / SHA-384
- ECDSA / SHA-384

Each vector includes the simulator-emitted `TPMS_ATTEST` and `TPMT_SIGNATURE`,
plus the certificate material needed by the verifier.

## Why

#320 landed canonical support for marshalled `TPMT_SIGNATURE` envelopes, RSASSA,
RSAPSS, ECDSA, and SHA-256/384/512.

Its unit coverage generates signatures using helpers living beside the verifier.
Those tests are valuable, but the producer and consumer still share the same
implementation context.

These fixtures add a different provenance boundary: the attestation and signature
bytes were emitted by the Microsoft/TCG TPM 2.0 reference implementation.

All four reference vectors verify successfully against the merged #320
implementation, both as marshalled `TPMT_SIGNATURE` bytes and as
`ParsedSignature` values.

The reference corpus also exercises RSAPSS/SHA-256, which the existing
parameterized SHA-384 fixture does not cover, plus metadata/key-type mismatch
branches and tamper rejection.

## Provenance

Reference implementation:

`microsoft/ms-tpm-20-ref`
commit `ee21db0a`
TPM revision 1.83

Environment used to capture the vectors:

- tpm2-tools 5.6-1build4
- tpm2-tss 4.0.1

The `TPMS_ATTEST` and `TPMT_SIGNATURE` bytes are **REFERENCE_SIMULATOR_EVIDENCE**,
not physical-hardware captures.

The AK certificate chains are synthetic wrappers around the exact simulator AK
public keys and are documented as mixed provenance in the fixture README.

No private key material is committed.

## Test coverage

`python/tests/test_tpm_reference_vectors.py` adds 26 tests:

- each of the four vectors verifies as raw marshalled `TPMT_SIGNATURE` bytes
- each of the four verifies as a `ParsedSignature`
- a fixture self-description guard, so a renamed or regenerated file cannot
  quietly retarget the tests at a different algorithm pair
- a genuine signature with a wrong declared hash is rejected
- a genuine signature with a wrong declared scheme is rejected
- ECDSA-against-RSA-key and RSASSA/RSAPSS-against-EC-key both raise
- tampered `TPMS_ATTEST` fails, for all four vectors
- tampered signature fails, for all four vectors
- the reference PSS signature verifies under digest-length salt semantics,
  asserted against the signature rather than against the verifier

Measured on this branch:

- targeted TPM tests (`test_tpm_verify.py`, `test_tpm_hardware_vector.py`,
  `test_tpm_reference_vectors.py`): 70 passed
- full suite: **1188 passed, 6 skipped** (1162 passed, 6 skipped on `main`; +26,
  no existing test modified)
- `mypy src/agent_manifest`: no issues in 24 source files
- `ruff check src/ tests/ --select E,F,W --ignore E501`: passes
- `bandit -r src/ -c pyproject.toml`: 0 issues
- `git diff --check`: clean

## Scope

No verifier behavior is changed by this PR.

This is follow-up conformance/regression evidence for the TPM verification path
implemented by #320.

Refs #255.

## DCO

All commits are signed off under the Developer Certificate of Origin.


imran-siddique:
This is conflicting with main and cannot be merged as it stands.

Could you rebase? I did not resolve it here: it changes TPM signature scheme and digest handling in `_tpm_verify.py`, and a resolution I guessed at in signature verification could weaken it without failing any test.

Everything else on the PR is green, and the reference vector fixtures are a welcome addition.

devdiv07:
Thanks for the steer. Rebased onto current main and took upstream `_tpm_verify.py` wholesale rather than attempting a conflict merge.

The PR is now test/fixture-only: zero production-code changes. I reframed it around the Microsoft/TCG reference-simulator vectors, and all four combinations (RSASSA/SHA-256, RSAPSS/SHA-256, RSASSA/SHA-384, ECDSA/SHA-384) verify against the #320 implementation in both marshalled-envelope and `ParsedSignature` forms.

I also removed the old tests that contradicted the canonical #320 semantics (SHA-512-as-unsupported and no-envelope-sniffing) and kept only the reference-provenance and negative coverage.

Full suite: 1188 passed, 6 skipped (1162/6 on main; +26, no existing test modified).
mypy / CI ruff / bandit / `git diff --check` all pass.

imran-siddique (approved):
Thanks for taking upstream `_tpm_verify.py` wholesale rather than resolving the conflict. That was the right call for exactly the reason I did not resolve it myself: a guessed merge inside signature verification can weaken it without failing a single test, and now there is nothing to guess at. The PR is test and fixture only.

The value here is the provenance boundary, not the vector count. `test_tpm_verify.py` generates its signatures with a helper living beside the verifier that reads them, so producer and consumer share one interpretation of the wire format, and a shared misreading of `TPMT_SIGNATURE` would pass. These bytes came from the reference implementation instead, which is the first thing in this repo's TPM coverage that the repo did not also produce. The RSAPSS/SHA-256 combination is a genuine gap closed as well; the in-repo parameterization only reaches that scheme at SHA-384.

I verified the provenance rather than taking it from the README. `microsoft/ms-tpm-20-ref` commit `ee21db0a941decd3cac67925ea3310873af60ab3` exists, is dated 2024-10-04, and its commit message is "Integrate TPM 2.0 v183 changes from Trusted Computing Group", which matches the claimed revision 1.83 and the `TPM2_PT_MANUFACTURER` of `"XYZ "`. CI ran on the rebased head: 12 checks green, so the only red mark is the maintainer hold.

The fixture README is the part I would keep as the pattern for future evidence corpora in this repo. It classifies the corpus `REFERENCE_SIMULATOR_EVIDENCE`, says plainly that these are not real hardware, discloses that the AK chains are synthetic wrappers around the simulator's own public keys rather than letting a reader assume a real EK-rooted chain, and states the limit that actually matters: a vendor part may never emit RSAPSS or SHA-384 quotes at all, so what these establish is that a quote of that shape verifies, and nothing about any fleet. Naming `test_tpm_hardware_vector.py` as the separate and stronger claim keeps the two from being read as one.

That distinction is the whole reason this repo's attestation claims hold up, and I would rather have a corpus that states its own ceiling than one that quietly reads as hardware evidence.

Merging.

Issue #325, opened 2026-08-20
spec_change: key-to-issuer authorization has no normative resolution path (§3.6, §5.3)


litzki-systems:
## Spec section

Sections 3.6, 5.3, 3.7, 3.8. Review tracks 4 and 5 from ws4-secure-design-agentic-systems#149.

## Problem

PR #220 closed Issue #215 at the SDK level. `verify_manifest` now resolves the signing key and checks that the key is authorized for the declared issuer before signature verification begins. The specification does not require this. A conforming implementation that skips the check is conforming.

Section 3.6 places the signature object outside the signing pre-image. `signature.key_id` is unauthenticated metadata. It identifies a key. It does not authorize one. OCSF's `authority_uid` under `record_integrity` covers the verification side of this: a verifier can confirm a signing credential belongs to the expected authority. What remains open is the normative prohibition the SDK enforces but the spec does not state.

Nothing prevents a verifier from accepting issuer authorization derived from an identifier the attested subject controls. Otherwise issuer and subject collapse into one role on every self-issued manifest. PR #220 enforces the separation via opt-in `trusted_key_issuers` with an empty-map default. Imran's review comment on #220: "Worth being explicit in the spec/README that issuer binding is only enforced when the map is populated."

This gap is compounded by #322: if the signing pre-image is computed on non-conformant canonical bytes, key authorization checks operate on a signature whose integrity is already in question.

Separately, the verification result reports VALID or MISMATCH but carries no signal about which trust path authorized the issuer key. The spec already marks weaker claims explicitly (PROVIDER_ASSERTED, UNVERIFIABLE). The issuer trust path has no equivalent.

## Proposed change

Two normative prohibitions:

- A verifier MUST accept only a key resolved through the defined authorization path. A key carried inside the manifest document is not a valid source.
- A verifier MUST reject issuer authorization derived from an identifier controlled by the attested subject.

Section 3.8 already traverses rotation chains. The transparency log, required at production level, can record key-to-issuer bindings through the same infrastructure.

Negative conformance vectors to make this testable: a key taken from the document itself, a key published under the subject's identity label, a valid signature from a key not authorized for the declared issuer, and a rotation chain with a gap.

## Impact

- Backward compatible: yes — PR #220 already enforces this in the SDK via opt-in
- Affects conformance level(s): all levels that verify a signed manifest
- Conformance tests that need updating: negative vectors needed (none exist today)
- Regulatory mapping impact: none — this narrows what a verification result means

[...]

Issue #322, opened 2026-08-19
agent-manifest rejects a TRACE record that trace-spec signed: the canonical form is not RFC 8785


lywinged:
## SDK version

`0.11.0` (`main` at `2ce6c97`)

## Python version

`3.11.15`

## What happened

`_trace.py:210` computes the TRACE signature pre-image with this package's `canonicalize()`:

```python
return canonicalize({k: v for k, v in envelope.items() if k != "signature"})
```

`canonicalize()` is not RFC 8785 conformant. The clearest case is object key ordering — `_serialize_dict`, `_canonicalize.py:100-105`:

```python
def _serialize_dict(d: dict[str, Any], *, exclude_none: bool, depth: int) -> str:
    # RFC 8785 §3.2.3: sort keys by Unicode code point order.
    # Python's str comparison uses Unicode code point order by default — no
    # special locale or collation needed.
    parts: list[str] = []
    for k in sorted(d.keys()):
```

RFC 8785 §3.2.3 sorts by **UTF-16 code unit**, not code point. The two agree across the BMP and part company once a key holds a supplementary-plane character, because its surrogate pair sorts below every high BMP character while its code point sorts above them.

trace-spec names this trap by name. `spec/trace-v0.2.md` §3.2.2:

> Object keys are sorted by UTF-16 code unit (ascending), per RFC 8785 §3.2.3. [...] Sorting Python `str` values with `sorted()` gives code-point order and is wrong here; an RFC 8785 library gets this right.

and, in the same section:

> Implementations MUST use an RFC 8785-conformant library.

## What you expected

A TRACE record signed by trace-spec's own tooling verifies here. It does not.

Running trace-spec's published canonicalization-boundary corpus (`examples/canonicalization-boundary/`, four records, each valid against its own declared key over its own RFC 8785 bytes) through `trace_signing_pre_image`:

```
vector                     pre-image bytes   trace-spec verify   agent-manifest verify
non-ascii-values           identical         VALID               VALID
non-bmp-values             identical         VALID               VALID
utf16-key-order            DIFFERENT         VALID               REJECTED
utf16-key-order-nested     DIFFERENT         VALID               REJECTED

trace-spec's own verifier accepts   4/4
agent-manifest's pre-image accepts  2/4
```

Same records, same keys, same signatures.

Key order is not the only axis. A differential of `canonicalize()` against a reference JCS library also finds:

| | this SDK | RFC 8785 |
|---|---|---|
[...]

PR #329, opened 2026-08-21
docs: say that issuer key authorization is opt-in and off by default (#325)

imran-siddique:
Documentation only. No behaviour change, no normative requirement added.

## Why

`verify_manifest()` performs no issuer-key authorization unless `VerificationContext.trusted_key_issuers` is populated, and it defaults to an empty map. `_verify.py` opens the check with:

```python
if not trusted_key_issuers:
    return None
```

So out of the box, `signature.key_id` names a key and nothing confirms that key is authorized for the manifest's `issuer`. A `VALID` result from a default context means the signature verifies under a key the caller trusted, not that the key belongs to the issuer the manifest names. `signature.key_id` sits outside the signing pre-image (§3.6), which is what makes the distinction matter: it is unauthenticated metadata that identifies a key without authorizing one.

Neither the specification nor the SDK docs said this. This is the outstanding item from my own review of #220 ("worth being explicit in the spec/README that issuer binding is only enforced when the map is populated"), and #325 raised it again from the outside.

## What

- **§5.3**, a clearly-marked non-normative implementation note before the `challenge_nonce` paragraph: what a `VALID` result does and does not cover here, that the reference SDK's behaviour is not a substitute for a requirement, and that the normative question is open and tracked in #325.
- **`docs/tutorials/server-side-verification.md`**, beside the existing `trusted_key_issuers` mention, so the default is stated where someone configuring a verifier will actually meet it.

Both point at #325 rather than pre-empting it. Whether a verifier MUST reject a key not resolved through a defined authorization path, and MUST reject issuer authorization derived from an identifier the attested subject controls, is normative text and goes through the spec change process with a sponsor. This PR only stops the current behaviour from being undocumented while that runs.

## Note

#325 also observes that this compounds with #322: if the signing pre-image is computed on non-conformant canonical bytes, key authorization is being applied to a signature whose integrity is already in question. #322 is confirmed and its fix is being prepared separately.


Qiang-Xu (approved):
It makes sense to me, thanks!

PR #331, opened 2026-08-23
security: close integrated verification trust gaps

imran-siddique:
Fixes the published GHSA-ww2p-prj4-c6xf and prepares agent-manifest 0.11.1.\n\nSecurity changes:\n- authenticate enforced HITL approvals and bind them to the current manifest\n- fail closed when bound runtime artifacts were not observed\n- require independently appraised, manifest-bound transparency evidence at Level 1+\n- reject forged or cross-manifest-replayed transparency receipts\n\nEvidence on the rebased candidate:\n- 1209 passed, 6 skipped\n- wheel and sdist built successfully\n- twine check passed\n- git diff --check passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,18 @@
   signatures, and approvals replayed from another manifest fail closed. See
   GHSA-ww2p-prj4-c6xf.
 
+- **[SECURITY][SDK]** Bound runtime artifacts now fail closed when their
+  observed hashes are absent. Signature-only appraisal remains available only
+  through an explicit API or CLI opt-out and must not be used as authorization
+  evidence about a deployed agent.
+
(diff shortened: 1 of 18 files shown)


zohebk8s (approved):
Approving. The HITL approval signature check and the fail closed artifact binding both look right.

One gap I hit while testing: `manifest verify` has no option that populates `approver_public_keys`. Whenever an approval is actually evaluated, meaning `hitl_record.required` is true or `--enforce-hitl` is passed, it resolves to UNVERIFIABLE, including approvals that are correctly signed and bound to the manifest being verified. So `--enforce-hitl` has no input that can produce VALID. Raising a follow-up PR with `--approver-key` and the CLI HITL tests.

Two things for this PR: AM-VEC-009 signs its approval with the issuer key, and 0.11.1 has no changelog section of its own.

PR #332, opened 2026-08-23
fix(cli): add --approver-key so HITL verification can succeed

zohebk8s:
Follow up to #331, which closed GHSA-ww2p-prj4-c6xf on the integrated path.

## Problem

The integrated verification path now requires a trusted approver key for every HITL approval, but `manifest verify` has no option that populates `approver_public_keys`. Whenever an approval is actually evaluated, meaning `hitl_record.required` is true or `--enforce-hitl` is passed, it resolves to UNVERIFIABLE, including approvals that are correctly signed and bound to the manifest being verified. So `--enforce-hitl` has no input that can produce VALID.

Because `hitl_record.required` triggers the check on its own, this also reaches callers that never passed `--enforce-hitl`. A third party manifest declaring `hitl_record.required` makes any CLI verification UNVERIFIABLE. The practical risk is that operators learn to ignore UNVERIFIABLE.

Reproduced against the real console script, same three manifests on main before and after #331, issuer key supplied with `--public-key`, `--signature-only` added where it exists so the artifact binding change is not the variable:

```
case                                          before #331      after #331
correctly signed approval, --enforce-hitl     VALID/APPROVED   UNVERIFIABLE
forged approval signature, --enforce-hitl     VALID/APPROVED   UNVERIFIABLE
correctly signed approval, no --enforce-hitl  VALID/APPROVED   UNVERIFIABLE
```

Row two is the advisory and #331 correctly closes it. Row one is the problem: that approval is signed by a real approver key over the correct pre image and bound to the right manifest_id, and it still cannot pass.

## Change

`--approver-key APPROVER_ID=PATH`, repeatable, maps one `approver_id` to a raw Ed25519 public key hex file. The hex loading and 32 byte validation already used by `--public-key` are factored into a shared helper so both options validate identically.

A failing result now names the missing input. When `fields_verified.hitl_record` is UNVERIFIABLE, stderr points at `--approver-key` instead of leaving a bare UNVERIFIABLE.

`docs/api-reference/cli.md` is regenerated with `scripts/gen_cli_reference.py`. The explanation lives in the command docstring, which the generator picks up.

Adding the option does not weaken anything. The three cases above are unchanged without it, and row one only reaches VALID once a trusted approver key is supplied:

```
correctly signed approval, --enforce-hitl                    UNVERIFIABLE
forged approval signature, --enforce-hitl                    UNVERIFIABLE
correctly signed approval, no --enforce-hitl                 UNVERIFIABLE
correctly signed approval, --enforce-hitl --approver-key     VALID/APPROVED
```

## Tests

Adds the CLI HITL coverage that was absent both before and after #331. No test in `python/tests/test_cli.py` matched `hitl` on either commit, which is why the suite stayed green while the flag was unusable.

- genuine approval with `--approver-key` verifies
- forged approval signature is INVALID
- approval replayed from another manifest is INVALID
- missing approver key is UNVERIFIABLE and the hint names the option
- `hitl_record.required` reaches the path without `--enforce-hitl`
- malformed `APPROVER_ID=PATH` and missing key file fail cleanly

Full suite: 1216 passed, 6 skipped.

## Note

Scope: this covers only the CLI gap. The other two items raised in review on #331 are still open on main and are deliberately not touched here. AM-VEC-009 signs its approval with the issuer key, which the advisory lists as an indicator, and it is a published conformance vector other implementations build against. And `pyproject.toml` reads 0.11.1 while the changelog has no `## [0.11.1]` section, so a release cut today would ship without release notes under its own version. Both are worth closing before 0.11.1 is published.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@
   signatures, and approvals replayed from another manifest fail closed. See
   GHSA-ww2p-prj4-c6xf.
 
+- **[SECURITY][CLI]** `manifest verify` gained `--approver-key
+  APPROVER_ID=PATH` (repeatable) to supply trusted HITL approver keys. Without
+  it the CLI had no input that could populate `approver_public_keys`, so every
+  approval resolved to `UNVERIFIABLE` and `--enforce-hitl` could not succeed.
+  A failing result now names the missing option.
+
 - **[SECURITY][SDK]** Bound runtime artifacts now fail closed when their
   observed hashes are absent. Signature-only appraisal remains available only
   through an explicit API or CLI opt-out and must not be used as authorization
--- a/docs/api-reference/cli.md
+++ b/docs/api-reference/cli.md
@@ -112,8 +112,14 @@ Usage: manifest verify [OPTIONS] MANIFEST_FILE
 
   Use --crl-path to load a revocation list and check for revoked manifests.
 
+  HITL approvals attach outside the manifest signature, so supply the
+  approver keys you trust with --approver-key. Without them an approval is
+  UNVERIFIABLE and the manifest can never verify.
+
   Example:
     manifest verify attested.json --crl-path revocations.jsonl
+    manifest verify signed.json --public-key pub.hex --enforce-hitl
+      --approver-key mailto:alice@example.com=alice.hex
 
 Options:
   --enforce-hitl                  Fail unless a required HITL approval is present and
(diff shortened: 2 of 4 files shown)


imran-siddique (approved):
Reviewed the CLI trust-input path and negative coverage. Approver keys are explicit relying-party inputs; malformed and duplicate mappings fail closed; forged and cross-manifest approvals remain invalid; and absent trust remains unverifiable. The vector key-separation and 0.11.1 release-note follow-ups remain separate pre-release work.

PR #338, opened 2026-08-24
fix: make canonicalize() RFC 8785 conformant (#322)

### LINKED ISSUE #322: agent-manifest rejects a TRACE record that trace-spec signed: the canonical form is not RFC 8785

SnehD:
## What

`canonicalize()` is the single RFC 8785 pre-image used for manifest signatures, COSE signing, delegation chains, revocation records, memory deltas, plugin bundles, and TRACE envelopes. It was not RFC 8785 conformant.

## Why

Fixes #322. Four conformance bugs, all in `_canonicalize.py`:

1. Sorted object keys by Unicode code point instead of UTF-16 code unit (RFC 8785 §3.2.3) — diverges when a key contains a supplementary-plane character.
2. Escaped U+007F, U+0080-U+009F, U+2028 and U+2029, which are outside ECMAScript JSON.stringify's escape set (RFC 8785 §3.2.2.2).
3. Float formatting used ad-hoc shortcuts instead of ECMAScript Number::toString (RFC 8785 §3.2.2.3): wrong integer/exponential cutover, wrong exponent padding, wrong small-magnitude threshold.
4. Silently serialized integers beyond the IEEE-754 safe range instead of refusing them.

## Spec impact

No normative text changed. Brings the implementation into conformance with the existing RFC 8785 requirement in Section 2.3. Conformance test AM-CRYPTO-001 is directly affected — this fix is required for that test to be meaningful.

## Fix

- UTF-16 code-unit sort key for object keys.
- Escape set corrected to quote, reverse solidus, and U+0000-U+001F only.
- Full ECMA-262 Number::toString algorithm via `Decimal(repr(f)).normalize()`.
- Integers with `abs(value) > 2**53-1` now raise `ValueError`.

## Test plan

- [x] `pytest tests/test_canonicalize.py -q` → 57 passed (removed the test asserting the old non-conformant escaping behavior; added tests for UTF-16 key order, the number-formatting table, and the safe-integer bound)
- [x] `pytest tests/interop/test_trace_canonicalization_boundary.py -v` → passed against 4 real signed vectors vendored from `agentrust-io/trace-spec` (fetched via `git clone`, not retyped)
- [x] Load-bearing check: reverted `_canonicalize.py` only, reran the guard — `03-utf16-key-order.json` failed exactly as reported in #322; restored, passes
- [x] Full suite: 1096 passed, 6 skipped, 0 failed (no regressions)
- [ ] `mypy src/agent_manifest` — not run in this environment
- [ ] `ruff check src/ tests/` — not run in this environment
- [ ] `CHANGELOG.md` — not yet updated

## Compatibility note

This changes canonical bytes for the affected value classes (supplementary-plane keys, values needing >2^53 or specific exponent ranges, and the four literal characters above). Anything already signed at one of those values stops verifying against itself under the corrected canonicalizer — though it was already unverifiable against any conformant implementation, which is the point.

## DCO

- [ ] Sign-off not yet added to these commits — will add before merge if required.


imran-siddique (changes requested):
This is careful work and three of the four fixes are right and worth landing. The fourth is not conformance, and I want to settle all of it in one pass rather than send you round twice.

## The three that are correct

UTF-16 code-unit key order, the escape set narrowed to quote, reverse solidus and U+0000-U+001F, and the full ECMA-262 `Number::toString` are genuine RFC 8785 requirements and the previous implementation was wrong on each. Vendoring trace-spec's four signed vectors as a cross-repository guard is the right instrument, and proving it load-bearing by reverting `_canonicalize.py` and watching `03-utf16-key-order.json` fail is what makes the claim checkable rather than plausible. Removing `test_line_separator_escaped` is correct too: U+2028 is a hazard when JSON is pasted into JavaScript source, not a JSON serialization rule, and that test was pinning the bug in place.

## The fourth: refusing integers past 2^53-1 is not RFC 8785

I went to the RFC rather than reasoning about it. Appendix B, note (1):

> For maximum compliance with the ECMAScript "JSON" object, values that are to be interpreted as true integers SHOULD be in the range -9007199254740991 to 9007199254740991. **However, how numbers are used in applications does not affect the JCS algorithm.**

Note (2) then puts 2^68 in the reference vector table specifically to say the algorithm does not treat it as special. So JCS serializes these; it does not refuse them. The comment in the diff, "RFC 8785 numbers must round-trip through a double", is not something the RFC says.

**And the implementation disagrees with itself on the RFC's own published value.** I ran your `_canonicalize.py` against exactly the Appendix B entry:

```
json.loads('{"v":295147905179352830000}')   ->  Python int
  canonicalize(...)  ->  ValueError: exceeds the IEEE-754 safe integer range

the same value as the double behind Appendix B's hex 4430000000000000
  canonicalize(...)  ->  b'{"v":295147905179352830000}'   (correct)
```

Same number from the spec's own test data, opposite outcomes, decided by which Python type it happened to arrive as. That inconsistency exists independently of whether you agree with me about the RFC.

The practical cost is not hypothetical either: a nanosecond epoch timestamp is about 1.79e18 today, so an ordinary field value now raises inside the single pre-image used for every signature in this SDK.

If you want a safe-range restriction as **policy**, that is a defensible position, and it belongs as an explicit opt-in with its own name and its own reason, not inside the function that claims to implement JCS. Please drop it from this PR.

## The ADR hunk is stale, which is why this conflicts

Both this and #339 say "These are the only two Merkle hash operations selected by this ADR." #335 landed on main yesterday and added a third, the composite policy sub-bundle leaf in Section 3.2.2, which carries raw digest bytes rather than a JSON descriptor. The sentence is now false and the file has moved under you. Please rebase and re-word against the three operations that exist.

The rest of that hunk, pointing Sections 3.2.3 and 3.2.5.1 at the normative definitions and scoping RFC 8785 to where those sections define JSON as a hash input, is good and I would keep it.

## What I would like back

The three conformance fixes, the vendored vectors, and the corrected ADR wording, rebased onto current main, with the integer refusal removed. That is mergeable and I will take it. Thank you for filing #322 as an issue first and asking about timing rather than opening this cold, that was the right instinct on a change at the signature layer.

imran-siddique (changes requested):
You took the ruling and the code is right. One mechanical problem is blocking it, and it is separable from everything you fixed.

## The canonicalization fix is verified correct

Ran it rather than reading it:

- `isinstance(obj, int): return str(obj)`. The refusal is gone.
- The RFC 8785 Appendix B value that raised yesterday now serializes: `canonicalize(json.loads('{"v":295147905179352830000}'))` returns `b'{"v":295147905179352830000}'`. That is the reference vector's own expected output.
- The three fixes that were always right survived: safe-range integers serialize unchanged, and UTF-16 key order still sorts U+10000 before U+FFFD.
- `tests/test_canonicalize.py` plus `tests/interop/test_trace_canonicalization_boundary.py`: **58 passed**.

Thank you for taking the correction on the integer domain without argument. Three of four right on a change to the signing pre-image is a good PR, and the fourth was subtle enough that it took reading Appendix B's notes rather than its body to settle.

## What is blocking: the ADR file has committed conflict markers

`docs/adr/0003-rfc9162-merkle-domain-separation.md` on this head contains **nine conflict-marker lines**, including a `<<<<<<< HEAD` nested inside another, from two separate merges (`a43da11` and `3bb3ee5a`). Both the two-operation and three-operation sentences are present, unresolved, alongside two copies of the leaf-data paragraph.

There is also a corrupted line 60, a half-paste of the intact line 66:

```
id`. Composite policy sub-bundles sorted by policy language identifer (`cear`, `rego`, `yaml-agt), per Section 3.2.2
```

Three typos and an unclosed backtick. And the header now lists the composite policy bundle twice, once as "Section 2.2.3".

## What I would like

**Revert `docs/adr/0003-rfc9162-merkle-domain-separation.md` to main and push.** That is the only change needed; everything else here is ready and I will merge it.

```
git checkout 02632d53a263 -- docs/adr/0003-rfc9162-merkle-domain-separation.md
```

The ADR wording is worth having and I still want it, as its own PR against current main. It is a docs change with no dependency on the canonicalization work, and keeping them together means a verified fix to the signing pre-image waits on a merge artifact in a markdown file.

For that separate PR, the substance we agreed: three operations rather than two, and Sections 3.2.2, 3.2.3 and 3.2.5.1 named as the normative definitions with RFC 8785 scoped to where those sections define JSON as a hash input. Your line about the composite sub-bundle leaf using raw digest bytes and no JSON canonicalization is a good addition and I would keep it.

One piece of cruft to sweep while you are in there: `_MAX_SAFE_INTEGER` at `_canonicalize.py:30` is now defined and never referenced.

devdiv07:
While checking this head for any residual RFC 8785 differences, I found one historical behavior that I’m not sure is still intended.

_quote() still applies unicodedata.normalize("NFC", s), and test_nfc_normalization pins precomposed and decomposed Unicode to the same canonical bytes.

I checked the history before raising it: #3 explicitly required “Unicode strings in NFC normalization”, so this looks like an intentional historical requirement rather than an implementation slip.

The current contract seems different, though. RFC 8785 §3.1 says parsed JSON string data MUST NOT be altered during subsequent serializations, and specifically says JCS does not apply Unicode normalization — implementations must preserve Unicode string data “as is.” ADR-0001/current spec require RFC 8785 for canonical JSON; the NFC rules I could find are for raw text-artifact hashing rather than JSON canonicalization.

I reproduced this through the public API on main (605946b) and this head (1043afb) using:

precomposed: U+00E9
decomposed:  U+0065 U+0301

The RFC 8785 reference preserves distinct bytes; agent_manifest.canonicalize() collapses them in both value and key positions.

For the decomposed value:

reference preimage:      7b2276223a2265cc81227d
agent-manifest preimage: 7b2276223a22c3a9227d

Signing the reference preimage verifies over those producer bytes but fails over Agent Manifest’s reconstructed bytes.

I also removed only the NFC-normalization line in a disposable copy of this head: the decomposed case then matched the reference, and the canonicalization suite changed from 57 passed to 56 passed / 1 failed — the only failure was test_nfc_normalization. The TRACE boundary guard added here still passed.

So the observed impact is a fail-closed interoperability/conformance case for non-NFC input, not a signature bypass.

Is #3’s NFC rule still an intentional Agent Manifest profile deviation, or should it be removed as part of the RFC 8785 conformance work here? If the deviation is intentional, it looks like the current spec/ADR may need to state that exception explicitly.

SnehD:
Thanks for the detailed investigation.
I agree that the remaining NFC normalization appears to be a pre-existing
behavior introduced in #3 and is separate from the RFC 8785 fixes addressed
in #322.
 Since removing it would change canonical bytes and could affect
compatibility with existing signatures, we'd appreciate maintainer guidance
on whether this behavior should remain as an intentional project-specific
deviation or be removed for strict RFC 8785 compliance.

imran-siddique:
@SnehD Update on this, and a change of shape rather than a change of verdict.

**Your canonicalization work is right.** My 25 August review already confirmed the integer domain was fixed and 58 tests passed. Nothing has changed about that. What has kept this PR open for four days is not the code, it is the nine committed conflict markers in `docs/adr/0003-rfc9162-merkle-domain-separation.md`.

In the meantime #352 arrived with the same two fixes, scoped to `_canonicalize.py` and its test file only. I ran both implementations against the RFC 8785 Appendix B table from the IEEE-754 hex values before deciding anything: **you both score 10 of 10, identically**, including `5e-324`, 2^68 as `295147905179352830000`, and both boundary pairs. You differ only in how you express the UTF-16 sort, yours building a code-unit tuple and theirs encoding to UTF-16BE. Both correct.

I have merged #352, because it was unblocked and because a conformance defect in the signing pre-image should not wait on an ADR file. That is not a judgement between the two implementations and it is worth saying plainly that **you reached this answer first, on 24 August**.

**What I would like from this PR is the part #352 does not carry**, which is the more valuable half and is entirely yours:

- `docs/adr/0003-rfc9162-merkle-domain-separation.md`, with the markers resolved
- `python/tests/interop/vectors/canonicalization-boundary/*.json`, all four
- `python/tests/interop/test_trace_canonicalization_boundary.py`

Rebase onto current `main` and drop `python/src/agent_manifest/_canonicalize.py` and `python/tests/test_canonicalize.py` from the diff; those two are now redundant and will conflict. The float and UTF-16 assertions in `test_canonicalize.py` are the ones that became duplicative. Your interop vectors are not duplicative and I want them: they are the only thing in either PR that pins the boundary as data rather than as assertions in a test body.

On the ADR itself, the substance we agreed still stands: three operations rather than two, Sections 3.2.2, 3.2.3 and 3.2.5.1 named as the normative definitions, and your line about the composite sub-bundle leaf using raw digest bytes with no JSON canonicalization. Rewriting the file from your intended text will be faster than untangling the nested markers.

`_MAX_SAFE_INTEGER` at `_canonicalize.py:30` is now gone from main with the merge, so that piece of cruft is handled.

devdiv07:
@imran-siddique  One small scope check before _canonicalize.py is dropped from #338. I may be missing a follow-up change, but on current main after #352, _quote() still has:

      elif cp <= 0x001F or 0x007F <= cp <= 0x009F or cp in (0x2028, 0x2029):

#338 is the change that narrows this to cp <= 0x001F, and your earlier review identified that escape-set correction as one of the RFC 8785 fixes.

So I just wanted to check whether dropping _canonicalize.py entirely from #338 would leave that part of the fix behind.

Separately, the NFC normalization discussed above also still appears to remain on current main:

        s = unicodedata.normalize("NFC", s)

Since SnehD already noted that the NFC behavior may be compatibility-sensitive and needs maintainer guidance, would the intended split be to keep the escape-set correction in #338, and handle the NFC question separately once the project decides whether #3 remains an intentional deviation?

Just checking the intended scope before the rebase.

devdiv07:
@SnehD heads-up on the rebased head aa8867e3 --> I think “drop these from the diff” may have landed as “delete the files.”

The new interop test in this PR imports:

    from agent_manifest._canonicalize import canonicalize

but _canonicalize.py is absent from the branch tree, so that test cannot collect.

Against current main, both files are showing as deletions:

     D  python/src/agent_manifest/_canonicalize.py
     D  python/tests/test_canonicalize.py

Main still has both of the #352 versions, so I think the intended state was to carry those files over unchanged and have them simply disappear from the PR diff, rather than delete them from the branch.

That would leave #338 carrying the ADR clarification plus the interop vectors/test, which I think matches the scope requested above.

Flagging it because the current head's interop test depends on a module that the same head removes.

imran-siddique:
@SnehD you went a different way than I asked on the ADR and I think you were right, so let me say that first.

I asked you to revert `docs/adr/0003-rfc9162-merkle-domain-separation.md` to main and ship the wording separately. You cleaned it in place instead. The result is smaller and more coherent than what I asked for: the nine conflict markers are gone, and the substance we agreed is intact, both the three-operations sentence and the paragraph naming Sections 3.2.2, 3.2.3 and 3.2.5.1 as the normative definitions with RFC 8785 scoped to where those sections define JSON as a hash input. I am taking it as it stands. `_MAX_SAFE_INTEGER` is swept too; I checked.

**First, something you could not have known.** Two of this PR's CI runs, `CI` and `CodeQL`, were sitting in `action_required` and had never executed. Nobody had any signal on this branch, including me when I reviewed it. I released them. `CodeQL` passed. `CI` did not, and the rest of this comment is what it found.

## Two lines in the ADR still need fixing

Neither is on main, so both would land as new damage.

**1. The header lists the composite policy bundle twice**, the second time under a section number that does not exist:

```
**Spec section**: Section 4.1.1, Section 3.2.2 (composite policy bundle), Section 2.2.3 (composite policy bundle), Section 3.2.3 (tool catalog hash), Section 3.2.5 (RAG corpus)
```

Delete `Section 2.2.3 (composite policy bundle), `. There is no Section 2.2.3.

**2. A half-pasted fragment sits in the Alternatives section**, between the BLAKE3 and flat-hash entries, where a blank line belongs:

```
id`. Composite policy sub-bundles sorted by policy language identifer (`cear`, `rego`, `yaml-agt), per Section 3.2.2
```

It starts mid-word, misspells "identifier", and leaves a backtick unclosed. Delete the line and restore the blank.

## The merge conflict is the interesting part, so please do not resolve it mechanically

`mypy` fails on `_canonicalize.py:193` with `Unsupported operand types for + ("Literal['n']" and "int")`, three times. That line is `n = exponent + k`, and `Decimal.as_tuple()` types its exponent as `int | Literal['n', 'N', 'F']` because those literals stand for NaN and Infinity in the decimal internals.

**That line is not in your branch and not on main.** Your head has 160 lines and does not contain it. Main has 219 and does not contain it. It exists only in `refs/pull/338/merge`, which is what CI type-checks. Git produced a float serializer that neither of us wrote.

Here is why. While this PR was open, main gained a `_digits_and_exponent(abs_repr: str) -> tuple[str, int]` helper that returns a properly typed integer exponent and carries a long comment about Python and ECMAScript choosing fixed versus exponential form at different thresholds, 1e16 against 1e21, and Python zero-padding single-digit exponents where ECMAScript never does. Your branch predates that and has the shorter version. The textual merge interleaved them into an inline `as_tuple()` call, which is why the type error appears in a region neither side touched deliberately.

So when you rebase onto current `main`: **in the float-serialization region, take main's version wholesale.** Its `_digits_and_exponent` helper supersedes what your branch has there, and the typing problem disappears with it because the helper returns a real `int`. Keep your side everywhere else. Your PR is about integer serialization and UTF-16 key ordering, and none of that is in contention.

Once the two ADR lines are fixed and the rebase is done, I will merge. The canonicalization work itself I already verified correct and I have not asked you to revisit any of it.

imran-siddique (commented):
Review/test follow-up on `aa8867e3dfb0`.

The current branch deletes `python/src/agent_manifest/_canonicalize.py` and `python/tests/test_canonicalize.py`. This reproduces the deletion issue already raised in the thread: `python -m pytest -q` cannot collect, with 43 errors beginning at `_delegation.py:18` (`ModuleNotFoundError: agent_manifest._canonicalize`). Restore the current main versions so they disappear from the diff instead of from the package. Resolve the merge conflict and the two outstanding ADR text defects (duplicate Section 2.2.3 and the half-pasted `id` fragment), then rerun CI. There is no CI test run on the current head; the maintainer gate alone cannot establish test health.

Full local `python -m pytest -q` (Windows, Python 3.12, existing review environment): **43 errors in 3.54s**.

Contribution requirement still outstanding: these commits have no `Signed-off-by:` trailer: `845215a4`, `1c86aefd`, `cb5fbcb8`, `b015dc16`, `92d933c7`, `aa8867e3`. Please have the respective commit authors add their own DCO sign-offs, including merge commits, then request review of the resulting head.

imran-siddique:
Closing this as superseded, and asking for part of it back, so let me be precise about which part and why.

The canonicalizer half is landed. #404 merged on 7 September and closes #322. I wrote it after this review round, built from a differential harness run against the `rfc8785` reference implementation over 30,000 randomised documents, and it fixed the escaping half plus two further divergences that surfaced once the `strict=True` xfail came out. Precedence is the only reason that one went first and not a judgement about this one. Your `_canonicalize.py` and `test_canonicalize.py` hunks now apply to a file that has moved underneath them, which is why this shows as conflicting.

The review round here was not wasted. Three of the four fixes in your original diff were right, and the out-of-range integer ruling we worked through is now recorded rather than re-argued: RFC 8785 Appendix B note (1) makes the safe-integer range a SHOULD that "does not affect the JCS algorithm", and note (2) carries 2^68 in the vector table, so a conformant canonicalizer serializes it rather than refusing it. That is a circulating misconception and pinning it down was worth the round trip.

**What I want back.** This PR also adds `python/tests/interop/` with four canonicalization-boundary vectors, a runner, and a README, plus a one-line ADR-0003 correction adding the missing Section 2.2.3 reference. **None of that is on main.** #404 touched three files and none of them was an interop vector, so closing this flat would lose the corpus.

It is worth more now than when you wrote it. trace-spec#306 merged today for exactly this reason: the boundary set there had four vectors that all expected acceptance, so an implementation that accepted everything satisfied the whole set, and the repository had that recorded against itself as a known one-directional gap. A cross-implementation vector set here has the same job, telling an agent-manifest canonicalizer apart from one that merely does not crash.

Please open a fresh PR against current main with just `python/tests/interop/` and the ADR line, and leave `_canonicalize.py` and `test_canonicalize.py` alone. It should be a small clean diff and I will take it. Worth checking your four vectors against the two negative controls in trace-spec#306 while you are there: a vector set that agrees under both the correct rule and the wrong one discriminates neither, and that is the trap that set fell into.

PR #339, opened 2026-08-24
docs: clarify ADR-0003 selects only two Merkle hash operations

### LINKED ISSUE #337: 4.1.1 enumerates two Merkle constructions while ADR-0003 now names four
## Spec section

Section 4.1.1 (Merkle Tree Domain Separation). Related: ADR-0003, Section 3.2.2,
Section 3.2.6, Section 3.2.7.

## Problem

Section 4.1.1 states:

> All Merkle tree constructions in this specification (corpus `merkle_root`, tool
> `catalog_hash`) MUST use the RFC 9162 / RFC 6962 domain-separated hashing
> convention

and its leaf bullet points `leaf_data` at "the relevant section (section 3.2.3
for catalog, section 3.2.5.1 for corpus)".

#335 closed the composite leaf-encoding gap and updated ADR-0003, whose header
now lists Section 3.2.2 alongside 4.1.1, 3.2.3 and 3.2.5. Section 4.1.1 was not
part of that change. The clause claiming to cover all constructions enumerates
two, the ADR names four, and a reader starting at 4.1.1 does not reach the
composite rule #335 made normative.

The memory root in 3.2.6 and the merkle-log in 3.2.7 sit outside both lists as
well. Both define tagged leaves locally, so they are covered in practice rather
than by the umbrella clause.

## Proposed change

Either extend the parenthetical and the `leaf_data` pointer to name every
construction, or drop the parenthetical so "all" means all and the per-section
definitions are the single source. The second is less to maintain the next time
a construction is added:

> All Merkle tree constructions in this specification MUST use the RFC 9162 /
> RFC 6962 domain-separated hashing convention:
>
> - Leaf hash: `SHA-256(0x00 || leaf_data)`, where `leaf_data` is defined by the
>   section specifying that construction.

## Impact

- Backward compatible: yes. No construction changes; this states the scope the
  clause already intends.
- Affects conformance level(s): all
- Conformance tests that need updating: unclear, and this is worth a second pair
  of eyes. ADR-0003 says implementations that omit the prefixes "will fail
  conformance test AM-BIND-015", and I could not locate that test. Searching the
  repository, AM-BIND-015 appears only in ADR-0003 itself.
  `python/tests/test_am_bind.py` carries 44 test functions, none of which
  reference a conformance ID and none of which touch the `0x00` or `0x01`
  prefixes, while CONTRIBUTING.md asks for IDs in test docstrings. AM-VEC ids do
  resolve, as files under `python/tests/vectors/`. If AM-BIND-015 exists
  somewhere I have not looked, pointing me at it answers this line.
- Regulatory mapping impact: none

## Alternatives considered

Leaving it as is. The ADR is operative for an implementer who finds it, but
4.1.1 is where the specification asserts the rule and it currently under-claims
its own scope.

Naming only 3.2.2. That closes today's gap and leaves 3.2.6 and 3.2.7 outside a
clause that says all, which is the same defect one construction later.

SnehD:
## What

Clarifies ADR-0003 so it states plainly that only two Merkle hash operations are approved, instead of reading like there might be four.

## Why

Fixes #337. The "Alternatives considered" section listed rejected constructions (plain concatenation, BLAKE3, flat hash) right next to the two approved operations without saying they were rejected, so a reader could mistake them for additional supported options. The ADR also carried old leaf-content wording that no longer matches the current specification.

## Spec impact

No normative text changed. The ADR now explicitly defers to:
- Section 4.1.1 — the shared Merkle construction (the two approved hash operations)
- Section 3.2.3 — tool catalog leaf data and ordering
- Section 3.2.5.1 — RAG corpus leaf data and ordering

No conformance test IDs are affected.

## Test plan

- [x] Documentation-only change, no code affected
- [x] `git diff --check` passes (no whitespace errors)
- [x] Confirmed the outdated leaf-description wording is removed
- [x] No Markdown diagnostics on the file

## DCO

- [x] Sign-off not yet added to this commit — will add before merge if required.

### DIFF
--- a/docs/adr/0003-rfc9162-merkle-domain-separation.md
+++ b/docs/adr/0003-rfc9162-merkle-domain-separation.md
@@ -15,8 +15,14 @@ Use the RFC 9162 (Certificate Transparency v2) Merkle tree construction with exp
 - Leaf nodes: `SHA-256(0x00 || leaf_data)`
 - Internal nodes: `SHA-256(0x01 || left_hash || right_hash)`
 
-Leaf data for tool entries: RFC 8785 canonical JSON of the tool descriptor (schema + description, sorted by tool name).  
-Leaf data for corpus documents: RFC 8785 canonical JSON of the document descriptor (hash + identifier + ingested_at).
+These are the only two Merkle hash operations selected by this ADR. The entries in
+"Alternatives considered" are rejected constructions, not additional supported
+operations.
+
+Section 4.1.1 of the specification is the normative definition of the shared
+construction. Sections 3.2.3 and 3.2.5.1 normatively define each artifact's leaf
+data and ordering; they take precedence over this ADR for those details. RFC 8785
+applies only where those sections define JSON as an input to a hash.
 
 ## Rationale
 


imran-siddique:
Closing this one, and the reason is not that the idea is wrong. It is that the premise moved.

**This diff is byte-identical to the ADR hunk in #338.** Same blob transition, `5b9bea6..851f3c4`, same eight lines. So it carries no content #338 does not already carry, and merging both would conflict them against each other.

**The claim itself is now false.** Both PRs say "These are the only two Merkle hash operations selected by this ADR." #335 merged to main yesterday and added a third: the composite policy sub-bundle leaf in Section 3.2.2, which carries the raw digest bytes of each sub-bundle hash rather than a JSON descriptor. That is why both PRs are showing conflicts against `docs/adr/0003-rfc9162-merkle-domain-separation.md`. The file on main now describes three operations, not two.

Unlike #338, there is no separable correct half here to keep, since the stale sentence is the whole PR.

The part of the hunk worth preserving is the second half, pointing Section 4.1.1 at the shared construction and Sections 3.2.3 and 3.2.5.1 at each artifact's leaf data, and scoping RFC 8785 to where those sections define JSON as an input to a hash. That is a real clarification and it survives the third operation existing. I have asked for it in the #338 review, rewritten against the three operations that are actually there, so the ADR gets touched once rather than twice.

Full detail in the #338 review, which also covers the canonicalization work. Thanks for splitting the docs change out in the first place, that instinct was right; it just happened to be the half that main moved under.

PR #345, opened 2026-08-26
fix(verify): cross-check enforcement_mode when verifying policy_bundle

### LINKED ISSUE #576: Agent Manifest binding does not cross-check cMCP enforcement_mode required by §6.2
## Problem

Agent Manifest v0.2 §6.2 requires the cMCP attestation `enforcement_mode` to match the Agent Manifest `artifacts.policy_bundle.enforcement_mode`, with a conflict treated as an attestation failure.

The current cMCP → Agent Manifest verification path appears to bind runtime policy and tool catalog hashes through `agent_manifest.verify_manifest()`, but does not pass or compare the runtime `enforcement_mode`.

This looks like the unfinished follow-up already noted in #342:

> Extend to cover enforcement_mode attestation binding (Claim 1, Property 3)

That issue was closed after the hash-binding experiment shipped, but the enforcement-mode follow-up itself does not appear to have been implemented.

## Reproduction / current path

The current Agent Manifest integration uses `VerificationContext` with values such as:

* `policy_bundle_hash`
* `tool_catalog_hash`
* trusted signing keys

but there is no corresponding runtime enforcement-mode value passed into the verifier.

At the same time, cMCP has runtime modes:

* `enforcing`
* `advisory`
* `silent`

while Agent Manifest defines:

* `enforce`
* `advisory`
* `audit-only`

and §6.2 requires the two sides to match.

Example mismatch:

```text
cMCP runtime enforcement_mode = advisory
Agent Manifest policy_bundle.enforcement_mode = enforce
```

The current binding path appears to still be able to verify the policy hash/catalog hash without explicitly rejecting this mode mismatch.

## Expected behavior

The cMCP → Agent Manifest verification path should cross-check the runtime/attested enforcement mode against `artifacts.policy_bundle.enforcement_mode`.

A mismatch should fail verification / attestation rather than returning a valid binding.

## Actual behavior

The canonical Agent Manifest verification integration checks the policy bundle hash and other runtime bindings, but does not appear to perform an enforcement-mode comparison.

## Impact

This is narrower than a policy-bundle hash bypass: the policy bundle itself can still match exactly.

The gap is that the same approved policy bundle could be present while the runtime is operating in a different enforcement mode than the manifest declares.

## Suggested narrow fix

One possible implementation would be:

1. Define the cMCP ↔ Agent Manifest mode mapping explicitly:

   * `enforcing` → `enforce`
   * `advisory` → `advisory`
   * `silent` → `audit-only` (if this is the intended semantic mapping)
2. Pass the normalized runtime mode into the Agent Manifest verification path.
3. Reject a mismatch with a non-VALID result.
4. Add a negative test for:

   * manifest `enforce`
   * runtime `advisory`
   * expected result: mismatch / attestation failure

Related:

* #342
* #323
* agentrust-io/agent-manifest#176
* agentrust-io/agent-manifest#178

rajnisht7:
## What

Makes verify_manifest actually cross-check policy_bundle.enforcement_mode against the runtime's attested mode, instead of only checking the policy_bundle hash.

## Why

Fixes:  [cmcp#576 (agentrust-io/cmcp).](https://github.com/agentrust-io/cmcp/issues/576)

Spec section 6.2 already says this is required: a verifying party MUST reject a manifest whose enforcement_mode doesn't match the runtime's attested mode, and the spec's own table lists it as "MUST match. Conflict = attestation failure." But the SDK's VerificationContext never had a place to give it the runtime's mode, and verify_manifest never checked it so a manifest requiring enforce could match on hash alone while the runtime was actually running advisory (warn-only) or silent, and verification would still say VALID. This PR implements the check the spec already requires.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO

### DIFF
--- a/python/src/agent_manifest/_verify.py
+++ b/python/src/agent_manifest/_verify.py
@@ -256,6 +256,12 @@ class VerificationContext(BaseModel):
 
     system_prompt_hash: Optional[str] = None
     policy_bundle_hash: Optional[str] = None
+    # Spec 6.2: the runtime's attested policy-enforcement mode ("enforce",
+    # "advisory", or "audit-only"). Compared against
+    # artifacts.policy_bundle.enforcement_mode when the manifest declares one.
+    # A caller whose runtime has no notion of enforcement mode simply leaves
+    # this unset, which skips the check (matches today's behavior).
+    enforcement_mode: Optional[str] = None
     tool_catalog_hash: Optional[str] = None
     model_version: Optional[str] = None
     rag_corpus_merkle_root: Optional[str] = None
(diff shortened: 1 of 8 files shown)


imran-siddique (approved):
Reviewed at 68ca1b2. The SDK now fails closed when a manifest declares policy_bundle.enforcement_mode but the runtime context omits it or supplies a different value; matching and legacy-absent cases remain compatible. The mismatch is surfaced as policy_bundle.enforcement_mode and forces the aggregate policy_bundle field to MISMATCH. Tests cover match, mismatch, omission, and legacy absence. No unresolved threads or blocking findings.

PR #347, opened 2026-08-26
docs: make the standards landscape visible and hands-on

imran-siddique:
## Summary

- replace the empty Mermaid output with a self-contained, accessible SVG that renders without client-side JavaScript
- use a vertical four-layer layout with readable labels and unambiguous reference flow
- add real hands-on paths: manifest tamper detection, an end-to-end Manifest/cMCP/TRACE example, the runnable demo suite, and the AgenTrust Marketplace
- add the standards landscape to the Integrations navigation

## Evidence

- confirmed the current production page renders `<div class="mermaid"></div>` with no SVG or text
- loaded the replacement asset directly from the pushed branch: one SVG, 31 text nodes, visible at full size
- visually reviewed the rendered SVG and corrected two ambiguous edge routes before opening this PR
- `git diff --check origin/main...HEAD` exits 0

## Limitation

The local environment does not contain Python or MkDocs, so the full site build was not run locally. CI is the build check for this change.

Follow-up to #346 and cosai-oasis/ws4-secure-design-agentic-systems#149.


Qiang-Xu (approved):
Looks good to me. On the diagram box 2, it'll be nicer if the two arrows from SPDX and platform attestation boxes can point to `agent manifest` box

PR #352, opened 2026-08-28
fix(canonicalize): match RFC 8785 float formatting and UTF-16 key order

rajnisht7:
## What

`_canonicalize.py` implements RFC 8785 (JCS), but has two issues: object keys are sorted by Python’s Unicode code-point order instead of the RFC-required UTF-16 order, and floats are formatted using Python’s `repr()` instead of ECMAScript’s number formatting. This can produce different canonical bytes from other RFC-compliant implementations.

## Why

These differences can cause the same manifest to produce different signatures, hashes, or Merkle leaves across SDKs. For example, a float like `1e-7` or a key containing a non-BMP character could make a valid manifest fail cross-implementation verification. The fix was checked against the RFC 8785 example and 200,000 random IEEE-754 doubles.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO

### DIFF
--- a/python/src/agent_manifest/_canonicalize.py
+++ b/python/src/agent_manifest/_canonicalize.py
@@ -24,7 +24,6 @@ import math
 import unicodedata
 from typing import Any
 
-
 _MAX_DEPTH = 64  # DOS-006: prevent RecursionError from deeply nested JSON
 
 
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
Both fixes are real and I verified them by running the code rather than reading it.

I ran your `_float_to_str` against the **RFC 8785 Appendix B** sample table directly from the IEEE-754 hex values. **10 of 10 exact**, including the ones that actually discriminate an implementation:

| IEEE hex | Required | Yours |
|---|---|---|
| `0000000000000001` | `5e-324` | matches |
| `7fefffffffffffff` | `1.7976931348623157e+308` | matches |
| `4430000000000000` | `295147905179352830000` | matches |
| `44b52d02c7e14af5` / `...af6` | `9.999999999999997e+22` / `1e+23` | matches |
| `3eb0c6f7a0b5ed8d` / `...8c` | `0.000001` / `9.999999999999997e-7` | matches |

The two Python-versus-ECMAScript divergences you name in the docstring are both real and both bite: `repr()` switches to exponential at 1e16 where ECMAScript holds fixed notation to 1e21, and `repr()` zero-pads to `1e-07` where ECMAScript writes `1e-7`. Either one changes the signing pre-image.

`sorted(key=lambda k: k.encode("utf-16-be"))` is correct for §3.2.3 and is the neat way to do it: lexicographic comparison of UTF-16BE bytes is exactly comparison of UTF-16 code units, so surrogate pairs sort below U+E000-U+FFFF without materialising a code-unit list.

Thirteen checks green. Approving and merging.

**One thing you should know, because it is not your fault and it affects what happens next.** #338 landed the same two fixes four days ago and is still open, blocked on committed conflict markers in an unrelated ADR file rather than on its code. I ran both implementations side by side against the same Appendix B table and they score identically, 10/10 each; you differ only in style. I am merging yours because it is unblocked and focused, and asking #338 to rebase down to the part yours does not carry: the ADR and its four interop vector files.

Nothing for you to do. Recording it so neither of you finds out at a merge conflict, and so it is on the record that @SnehD reached the same answer first.

PR #354, opened 2026-08-29
fix(attestation): reject expired certificates in hardware attestation chains

rajnisht7:
## What

Three hardware-attestation verifiers each carry their own, independent certificate-chain verification logic instead of using the shared, more rigorous verify_cert_chain in _cert_chain.py:

_snp_verify.py -> verify_vcek_chain (AMD SEV-SNP: VCEK <- ASK <- ARK)
_tdx_verify.py -> verify_tdx_quote (Intel TDX: PCK leaf <- intermediate <- Intel SGX Root CA)
_tpm_verify.py -> _verify_ak_chain (TPM 2.0: AK <- vendor root)

All three checked signatures only. None of them checked that any certificate in the chain was within its validity period (not_valid_before / not_valid_after) the single most basic check any X.509 relying party is expected to perform (RFC 5280). An expired-but-otherwise-cryptographically-valid certificate at any position in any of the three chains verified successfully.

_cert_chain.py's own module docstring already claims: "the AMD-specific _snp_verify.verify_vcek_chain is a thin specialisation of it [the generic verify_cert_chain]". In the code as it stood, that was not true grepping all three verifier files for any reference to _cert_chain returns nothing; none of them imported from or called into it at all.

## Why

An attestation certificate's validity window is one of the mechanisms that bounds how long a given hardware/firmware/microcode state stays trusted AMD, Intel, and TPM vendors reissue these certificates (VCEK is explicitly re-issued per reported TCB/security-patch-level on AMD's KDS) precisely so that certificates tied to a since-superseded state stop being accepted. Skipping the validity-period check means a report or quote signed under an old, expired certificate including one whose corresponding key material or platform state is no longer considered trustworthy is still accepted indefinitely, with no time bound at all.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO

### DIFF
--- a/python/src/agent_manifest/_cert_chain.py
+++ b/python/src/agent_manifest/_cert_chain.py
@@ -19,8 +19,9 @@ each carrying their own chain verifier; the AMD-specific
 """
 from __future__ import annotations
 
+from collections.abc import Sequence
 from datetime import datetime, timezone
-from typing import TYPE_CHECKING, Optional, Sequence
+from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from cryptography import x509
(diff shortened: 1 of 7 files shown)


imran-siddique (approved):
Verified against `main` at `5d19b1d0042d` rather than taking the report, because this is a claim about the attestation layer:

```
_snp_verify.py   _cert_chain refs: 1   not_valid_before/after: 0
_tdx_verify.py   _cert_chain refs: 0   not_valid_before/after: 0
_tpm_verify.py   _cert_chain refs: 0   not_valid_before/after: 0
```

Three hardware-attestation chain verifiers, zero validity-period checks between them. An expired but cryptographically intact certificate at any position in a SEV-SNP VCEK chain, an Intel TDX PCK chain or a TPM AK chain verified successfully. That is the check RFC 5280 assumes every relying party performs, and it was absent from the layer Level 1 conformance rests on.

**The docstring finding is the more uncomfortable half.** `_cert_chain.py` opens with:

> This is the shared primitive the org's hardware-attestation verifiers build on.

Two of the three never referenced it. A module asserting it is the shared implementation, while the three consumers each carry their own, is worse than three independent implementations, because a reader auditing the attestation path finds the rigorous one, believes it is in use, and stops looking.

**What decided the merge for me is that `verification_time` is injectable.**

```python
verification_time: datetime | None = None,
...
now = verification_time or datetime.now(timezone.utc)
```

A certificate valid when the quote was produced and expired today is not the same as a certificate that was never valid, and collapsing them would have made this change break historical verification of records that were sound when issued. That is the historical-versus-present-tense distinction I have ruled on three times this week in other repos, and you got it right without it being raised. Rejecting naive datetimes explicitly, and reporting the window alongside the time it was checked at, are both the right calls: an expiry error that does not say *when you checked* is unactionable.

Tests for all three verifiers, thirteen checks, only the maintainer hold red.

Merging. The wider problem, three verifiers reimplementing chain logic beside an unused shared primitive that claims otherwise, is worth an issue on its own; this PR fixes the security consequence and leaves the duplication.

Issue #348, opened 2026-08-27
Section 3.5: hitl_record cannot express whether an approval is still admissible


ioanavalea:
## Spec section

This concerns Section 3.5, Human-in-the-Loop Approval Records, together with ADR-0006 and the `ROADMAP.md` v0.2 item "HITL approval scheme clarification".

## Problem

Although Agent Manifest can prove that the agent running now is the agent that was approved, it cannot yet express whether that approval is still admissible.

The shipped `HitlApproval` model records who approved (`approver_id`, `approver_identity_type`, `approver_role`), when (`approved_at`), what was covered (`approved_scope`, carrying artifacts, risk tier, conditions and `approval_duration_seconds`), how the signature was produced (`approval_method`), a signature over `{manifest_id, approved_at, approved_scope, approver_id}`, and an `evidence_uri`. ADR-0006 states that a manifest with a HITL requirement is `APPROVED` only when the approver appears in the verifier's allowed set, `approved_at + approval_duration_seconds > now`, the evidence binding matches, and the signature verifies against the approver's published key.

All four conditions can hold even after the decision the approval represents has stopped being defensible, because the record captures that a human authorised something at a point in time without capturing what that human was relying on, whether they can still stop the system, or whether anyone objected.

Time-bounding is a proxy for currency, and a weak one in both directions, since an approval does not expire when the material facts change on day three, whereas it does expire when nothing has changed on day ninety. The result is a manifest that verifies cleanly, with its chain intact, its signature good and its clock unexpired, while the reasoning that authorised the deployment no longer holds, which is a failure mode that survives correct cryptography.

The repository already contains the better pattern in one place, because `docs/memory-governance.md` defines TTL-based re-approval for the memory baseline and emits `MEMORY_DRIFT_DETECTED` when state moves without re-approval, which amounts to drift detection against a declared baseline. That pattern currently covers a single artifact, whereas the authorising decision, which is what all ten bound artifacts exist to make provable, has no equivalent.

## Proposed change

Three optional field groups on `hitl_record`, each with a corresponding member added to `HitlResult`, composing with the existing `--enforce-hitl` gate rather than replacing it.

[...]

PR #355, opened 2026-08-30
docs(spec): §3.5 clarify that approval_duration_seconds bounds the window, not the currency, of an approval

ioanavalea:
## What

Adds one paragraph to §3.5 stating that `approval_duration_seconds` bounds the window in which an approval may be presented, not the period for which its grounds remain sound, and that a `VALID` result is not proof that those grounds still obtain.

## Why

§3.5 defines `approval_duration_seconds` as a required positive integer without saying what it bounds. Verification checks `approved_at + approval_duration_seconds > now`, so a reader can reasonably infer that an unexpired approval is a sound one. Expiry and admissibility are different questions, and the spec does not currently say so.

Follows the maintainer ruling in #348, which listed this among the items that can move without an organisational sponsor on the grounds that it states what is already true.

The wording deliberately mirrors the existing `risk_tier` caveat a few paragraphs above, which already establishes this pattern in the same section: a verifier proves only what was signed, and a `VALID` result MUST NOT be represented as proof that the declaration was correct.

## Spec impact

Section changed: §3.5 Human-in-the-Loop Approval Records, v0.2.

Normative delta: none. No RFC 2119 keyword is added that binds a producer. The single MUST NOT is a constraint on how a verification result may be represented, which restates the existing `risk_tier` treatment rather than introducing a new obligation.

Conformance test IDs: none. No existing vector changes behaviour and no new vector is required.

The same clarification would hold for v0.1 §3.5. I have left that out to keep the diff to one file, and can add it if preferred.

## Test plan

- [ ] `pytest -v` passes
- [ ] `mypy src/agent_manifest` passes
- [ ] `ruff check src/ tests/` passes
- [ ] New or updated tests cover the change

Documentation only, so the code checks above do not apply and no tests are affected. On the last item: this touches spec prose without a normative delta, so I have not added a CHANGELOG entry. Happy to add one if you would rather every spec-file edit be recorded there.

## DCO

Drafting and repository inspection were AI-assisted. §3.5, the `risk_tier` paragraph, and the verification behaviour in `_verify.py` were read against current main before writing.

### DIFF
--- a/spec/agent-manifest-spec-v0.2.md
+++ b/spec/agent-manifest-spec-v0.2.md
@@ -1026,6 +1026,8 @@ If the permitted actions span tiers, the highest tier applies. If reversibility 
 
 At Level 2, a verifier MUST return `APPROVAL_INSUFFICIENT` when any otherwise-current approval uses `software-key`, or when a `high` or `critical` approval uses anything other than `hardware-key`. The overall result MUST be `MISMATCH`. Level 0 records the declared method without making a regulatory-strength claim; Level 1 deployments define their minimum method in relying-party policy.
 
+`approval_duration_seconds` bounds the window in which an approval may be presented, not the period for which the reasoning behind it remains sound. Expiry and admissibility are separate questions: an approval can run out while nothing it relied on has changed, and it can remain unexpired after the basis on which it was given has ceased to hold. A verifier evaluating `approved_at + approval_duration_seconds > now` establishes only that the approval is within its declared window. It does not establish that the approval is still applicable, and a `VALID` result therefore MUST NOT be represented as proof that the grounds for the approval still obtain.
+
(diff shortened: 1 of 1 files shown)


imran-siddique (approved):
Exactly the clarification I said could move without a sponsor on #348, and it lands the distinction the thread spent three days establishing.

`approval_duration_seconds` reads as a currency bound until someone tells you it is not one, and the two failure directions you name are both real: an approval expiring while nothing it relied on has changed, and an approval remaining unexpired after its basis has ceased to hold. Stating both is what stops a reader treating expiry as a proxy for admissibility in either direction.

This states a property of the existing field rather than adding an obligation, which is why it needed no sponsor. The `approval_basis` field and any new `HitlResult` value remain held on #348.

Thank you for withdrawing your own open question 4 once @solloek369-arch pointed at the `ConfigurationAssurance` precedent. That saved a round and it is why this one is a two-line change rather than a redesign.

Issue #236, opened 2026-07-22
Azure TDX attestation via Azure MAA (deferred)


imran-siddique:
Deferred follow-up. **Azure TDX cannot be attested offline** (hardware-confirmed on Standard_EC2es_v6, westeurope): it runs behind the Hyper-V paravisor, exposes no `/dev/tdx-guest` and no configfs-TSM `tdx_guest` provider (driver won't bind), and surfaces only a **MAC'd `TDREPORT`** via the vTPM/HCL — not a signed DCAP quote. Rooting it as genuine silicon requires **Azure MAA** (networked) or an on-platform Quoting Enclave.

Offline TDX is already supported on non-paravisor guests (GCP C3, validated). On Azure, use SEV-SNP (`AzureCVMProvider`).

## If pursued
- Add an Azure-TDX collector (vTPM AK quote, mirroring `AzureCVMProvider`).
- Add a verifier that submits the TDREPORT/evidence to Azure MAA and validates MAA's signed result against MAA's published keys.
- This is a **networked** trust model (trusting MAA), distinct from the offline SNP/TDX paths — a deliberate product decision before implementation.

Limitation documented in LIMITATIONS.md (#235). Related: this closes the Azure half of the original TDX work; direct/offline TDX is done.

imran-siddique:
Closing this as not planned. It was filed as a deferral record and nothing has changed since.

Azure TDX still has no offline path (paravisor, no `/dev/tdx-guest`, no configfs-TSM provider, only a MAC'd TDREPORT via the vTPM). SEV-SNP via `AzureCVMProvider` covers Azure, and offline TDX covers non-paravisor guests, validated on GCP C3. Adding a networked, MAA-dependent verifier with no adopter asking for it is not worth an open issue.

The finding stays in the history here and in cmcp#409. Reopen when someone actually needs Azure TDX.

PR #357, opened 2026-08-30
fix(delegation): validate structure before interpretation

### LINKED ISSUE #356: verify_delegation_chain reads the root before structural validation
## Current behavior

`verify_delegation_chain()` documents malformed chains as `ValueError`, and `_validate_hop_structure()` says it is called before cryptographic verification so structural errors surface as `ValueError` rather than `IndexError` / `KeyError`.

At current `main` (`eb747f5fd610a8d7fa360e52faaea5db578e6b34`), the root is nevertheless read before that validator runs:

```python
if manifest_issuer:
    root = delegation_chain[0]
    root_identities = {
        root.get("principal_id"),
        root.get("principal_manifest_id"),
    }
    ...

root_max_depth = delegation_chain[0]["scope_grant"].get(
    "max_delegation_depth", DEFAULT_MAX_DELEGATION_DEPTH
)
...

for i, hop in enumerate(delegation_chain):
    _validate_hop_structure(hop, i)
```

That creates several exception edges before the documented structural boundary:

- a non-object root with `manifest_issuer` supplied reaches `root.get(...)` and raises `AttributeError`;
- an object root missing `scope_grant` raises `KeyError` before `_validate_hop_structure()` can report the missing required field;
- a root whose `scope_grant` is present but not an object reaches `.get(...)` and raises `AttributeError`.

There is a second part to the same boundary: `_validate_hop_structure()` itself assumes `hop` is an object (`hop.keys()`) and currently checks that `scope_grant` exists but not that it is an object. So simply moving the existing call upward does not close every malformed-root case.

## Expected behavior

Every malformed chain structure reaching this public verifier should fail through its documented `ValueError` boundary before root binding, depth evaluation, signature verification, or scope narrowing interprets nested fields.

A malformed JSON shape should not leak host-language `AttributeError` / `KeyError` to a caller that is written against the verifier's documented contract.

## Suggested narrow direction

Strengthen the existing structural authority rather than add a second validator:

1. `_validate_hop_structure()` first establishes that each hop is an object;
2. it establishes that `scope_grant` is an object before any caller uses `.get()` on it;
3. `verify_delegation_chain()` runs structural validation before the root-binding and root-depth reads that depend on those facts.

Whether validation is done as a complete first pass over the chain or as an equivalent ordering is an implementation choice; the invariant is that no structure-dependent read happens before the corresponding shape has been established.

## Regression matrix

At minimum:

- non-object root (`"bad"`, list, integer) -> `ValueError`, never `AttributeError`;
- root object missing `scope_grant` -> `ValueError`, never `KeyError`;
- root `scope_grant` = string/list/bool -> `ValueError`, never `AttributeError`;
- malformed later hop -> `ValueError` under the same structural boundary;
- valid single-hop and multi-hop controls unchanged;
- root/manifest-issuer mismatch still reports the existing binding failure once the root is structurally valid.

No schema, wire-format, delegation semantics, or scope-narrowing rule change is required. This is exception-safety and enforcement of the verifier's existing structural prerequisite.

## Adjacent observation, intentionally not included

`verify_hitl_approval()` in the same module has a similar nested-shape assumption around `approval.get("approved_scope", {}).get(...)`. I am leaving that out of this issue so this delegation-chain fix remains independently reviewable; I can file it separately after checking its call-path contract.

altrudev:
## What

Closes #356.

`verify_delegation_chain()` now establishes structural prerequisites before each part of a chain is interpreted:

- the root is validated before root/manifest binding and root-depth reads;
- each later hop is validated when verification reaches that hop, before that hop's signature or scope is interpreted.

The existing `_validate_hop_structure()` remains the verifier-side structural authority. It now establishes the object/primitive shapes that downstream verifier code assumes before using `.get()`, set construction, numeric comparison, or signature decoding.

For integer-valued scope fields, the guard follows JSON/JSON-Schema semantics rather than Python's host-language type hierarchy: booleans are refused, ordinary integers are accepted, and zero-fraction numeric values such as `3.0` / `3600.0` remain accepted. No parsed/normalized model is substituted into the signed preimage.

## Why

The old ordering read the root before `_validate_hop_structure()` ran. A non-object root, a missing/non-object `scope_grant`, or other malformed nested values could therefore escape as incidental `AttributeError`, `KeyError`, or `TypeError` even though this public API documents malformed chains as `ValueError`.

The main verification engine catches `InvalidSignature` and `ValueError` around delegation verification, so those host-language exceptions could also escape its intended fail-closed boundary.

## Regression evidence

The public-API matrix covers:

- non-list chains and non-object roots;
- missing/non-object `scope_grant`;
- malformed `principal_type`, optional `principal_manifest_id`, collection-valued scope fields, depth/TTL values, and signatures that would otherwise reach host-language operations;
- boolean/fractional numeric refusal plus `3`/`3.0` and `3600`/`3600.0` equivalence controls;
- a malformed later hop reached only after a valid signed root and trusted root key;
- valid single-hop control;
- the existing structurally-valid root/issuer mismatch behavior.

The later-hop control deliberately supplies a valid root and trusted key so its structural boundary, rather than an earlier key/signature failure, is load-bearing.

## Scope / non-claims

No schema, wire-format, delegation-authority, signature-preimage, or scope-narrowing rule is changed. This is prerequisite ordering and exception-safety.

The implementation deliberately preserves sequential failure ordering: a malformed later hop does not pre-empt an earlier failure while interpreting the root.

This is separate from #353, which changes TTL narrowing semantics.

AI-assistance disclosure: ChatGPT assisted with source triage, adversarial matrix design, mutation-oriented review, and drafting. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/python/src/agent_manifest/_delegation.py
+++ b/python/src/agent_manifest/_delegation.py
@@ -233,11 +233,9 @@ def verify_delegation_chain(
     if not delegation_chain:
         return
 
-    # Structure is a prerequisite for every later interpretation of a hop.
-    # Establish it once for the complete chain before root binding, depth,
-    # signatures, or scope comparisons consume nested fields.
-    for i, hop in enumerate(delegation_chain):
-        _validate_hop_structure(hop, i)
+    # The root is interpreted before the verification loop, so establish its
+    # structure before the root-binding and depth reads below.
+    _validate_hop_structure(delegation_chain[0], 0)
 
     # Bind the chain root to the manifest's signing identity. The root hop
     # establishes the authority the rest of the chain narrows from, so it must
@@ -271,6 +269,9 @@ def verify_delegation_chain(
     prev_scope: Optional[dict[str, Any]] = None
 
     for i, hop in enumerate(delegation_chain):
+        if i:
+            _validate_hop_structure(hop, i)
+
         if hop.get("hop") != i:
             raise ValueError(f"Hop {i} has wrong hop index: {hop.get('hop')}")
 
--- a/python/src/agent_manifest/_delegation.py
+++ b/python/src/agent_manifest/_delegation.py
@@ -56,6 +56,15 @@ _REQUIRED_HOP_FIELDS = frozenset({
 _SCOPE_LIST_FIELDS = ("tools", "data_classifications", "constraints")
 
 
+def _is_json_integer(value: Any) -> bool:
+    """True for JSON-Schema integer values, excluding Python booleans."""
+    if isinstance(value, bool):
+        return False
+    if isinstance(value, int):
+        return True
+    return isinstance(value, float) and value.is_integer()
+
+
 def delegation_depth_exceeded(chain_length: int, root_max_depth: int) -> bool:
     """Single depth rule shared by the Pydantic models and this verifier.
 
@@ -132,16 +141,14 @@ def _validate_scope_structure(scope: Any, hop_index: int) -> None:
             )
 
     depth = scope.get("max_delegation_depth", DEFAULT_MAX_DELEGATION_DEPTH)
-    if isinstance(depth, bool) or not isinstance(depth, int) or depth < 0:
+    if not _is_json_integer(depth) or depth < 0:
         raise ValueError(
             f"Delegation hop {hop_index} scope_grant.max_delegation_depth "
             "must be a non-negative integer"
         )
 
     ttl = scope.get("ttl_seconds")
-    if ttl is not None and (
-        isinstance(ttl, bool) or not isinstance(ttl, int) or ttl < 1
-    ):
+    if ttl is not None and (not _is_json_integer(ttl) or ttl < 1):
         raise ValueError(
             f"Delegation hop {hop_index} scope_grant.ttl_seconds "
             "must be a positive integer or null"
--- a/python/src/agent_manifest/_delegation.py
+++ b/python/src/agent_manifest/_delegation.py
@@ -173,6 +173,13 @@ def _validate_hop_structure(hop: Any, hop_index: int) -> None:
             f"Delegation hop {hop_index} missing required fields: {sorted(missing)}"
         )
 
+    hop_value = hop["hop"]
+    if not _is_json_integer(hop_value):
+        raise ValueError(
+            f"Delegation hop {hop_index} hop must be a JSON integer, "
+            f"got {type(hop_value).__name__}"
+        )
+
     principal_type = hop["principal_type"]
     valid_principal_types = _valid_principal_types()
     if not isinstance(principal_type, str) or principal_type not in valid_principal_types:
@@ -186,6 +193,12 @@ def _validate_hop_structure(hop: Any, hop_index: int) -> None:
     if not isinstance(principal_id, str) or not principal_id.strip():
         raise ValueError(
             f"Delegation hop {hop_index} has empty or non-string principal_id"
+        )
+
+    delegated_at = hop["delegated_at"]
+    if not isinstance(delegated_at, str):
+        raise ValueError(
+            f"Delegation hop {hop_index} delegated_at must be a string"
         )
 
     principal_manifest_id = hop.get("principal_manifest_id")
--- a/python/src/agent_manifest/_delegation.py
+++ b/python/src/agent_manifest/_delegation.py
@@ -53,6 +53,7 @@ _REQUIRED_HOP_FIELDS = frozenset({
     "hop", "principal_id", "principal_type", "delegated_at",
     "scope_grant", "delegation_signature",
 })
+_SCOPE_LIST_FIELDS = ("tools", "data_classifications", "constraints")
 
 
 def delegation_depth_exceeded(chain_length: int, root_max_depth: int) -> bool:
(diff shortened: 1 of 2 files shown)


imran-siddique:
@altrudev A maintainer response to the whole batch rather than eight separate reviews, because the same three things apply across most of it and you should hear them once.

## The class of finding is real and I want it

A verifier that raises a host-language exception on malformed input, instead of returning a recorded finding, is a genuine defect and not a style preference. A relying party wraps verification in `except` for the documented error type; anything else escapes as an unhandled crash on input an attacker chose. @lywinged swept the same class through `trace-spec` in #236, which merged yesterday, and it found nine. You are working a real seam and every one of your PRs ships tests, which most do not.

## Your CI had never run until an hour ago, and five of eight are red

Your PRs were sitting with runs in `action_required` under first-time-contributor gating, reporting 2 checks where those repos run 12 to 25. I released 36 held runs across the batch. Results:

| PR | failing step |
|---|---|
| trace-spec#254 | `Type check` (mypy) |
| trace-spec#257 | `Test` |
| agent-manifest#359 | `Type check` (mypy) |
| cmcp#596 | `Lint` |
| cmcp#597 | `Test` |

trace-spec#254's is concrete and is a type error the change introduces:

```
src/agentrust_trace/content_marking.py:163: error: Argument 2 to "_digest" has
incompatible type "Any | None"; expected "str"  [arg-type]
```

This is not a criticism of your local process, because you could not see any of it. It does mean the batch is not reviewable as it stands: I would be spending eight reviews telling one person about their own CI.

## Two of them collide with each other

**trace-spec#258 and #252** both modify `src/agentrust_trace/provenance.py` and `tests/test_provenance_cnf_boundary.py`. Whichever lands first conflicts the other. They also both concern the `cnf` boundary, so they may be one change split in two.

## What I am asking for

1. **Fix the five red ones**, now that you can see them. `mypy`, `ruff` and `pytest` locally will reproduce all five.
2. **Sequence #258 and #252**, or fold them into one. If they are genuinely separate concerns, say which lands first and rebase the other onto it.
3. **Then tell me the order you want them reviewed in**, and I will work them individually from the top.

I am deliberately not merging the three currently-green ones ahead of that. Not because of their quality, but because agent-manifest#357 is 229 lines against a delegation verifier and I would rather review it as one of a sequence you have staged than as one of nine competing for the same attention.

## One thing worth saying plainly

Seventeen issues and nine pull requests in a single day, from an account with nothing merged yet, is more than this project can absorb, and the constraint is review capacity rather than your throughput. A contributor who lands three changes and builds a track record gets the fourth reviewed faster than one who opens twenty-six.

The findings look real. Give me a queue rather than a pile and they will land.

altrudev:
Thanks for the batch response. I have treated review capacity as a constraint and stopped opening new lanes while this queue is being cleaned up.

**Requested review order:**

1. **trace-spec#252** — first. It is green on released CI and now independently approved by @lywinged.
2. **trace-spec#254** — second, after its fresh head executes. I fixed the mypy failure using the reviewed `isinstance(alg, str)` narrowing and added the unhashable near-miss matrix.
3. **agent-manifest#359** — third, after its fresh head executes. The released mypy failure was a single missing annotation and is fixed.

I will not ask you to review the rest until those three move.

**Sequencing / held work:**

- **trace-spec#258 stays Draft behind #252.** #252 lands first; then I will rebase #258 onto that exact new main and remove the stacked predecessor from its diff before marking it ready.
- **trace-spec#257**: the released test regression was fixed by restoring hash-mismatch precedence; the current head is held for later review.
- **agent-manifest#357**: currently green, but held behind the three-item queue above as you requested.
- **cmcp#596/#597** remain Draft reproducers and are not in the review queue. Their known host-language escapes are now narrowly xfailed so the draft itself does not poison CI; the desired result assertions remain active whenever the verifier returns normally.
- **agent-manifest#370** is also held behind this queue; I will not ask for review on it now.

I am not manually re-running held workflows or calling any fresh head green until the repository executes it. This should reduce the active review request to a queue rather than a pile.

carloshvp (changes requested):
The root ordering fix is sound, and I confirmed the full suite passes (`1275 passed, 6 skipped`). There is still one blocking gap in the public structural boundary this PR establishes.

`_validate_hop_structure()` does not establish the primitive types of two required fields before they are interpreted:

1. `hop` is only compared with `i`. In Python, `False == 0`, so a root containing JSON `"hop": false` passes the sequential-index check. Verification then constructs the signed preimage with `hop=i`, meaning the verifier checks integer `0` even though the record contains boolean `false`.
2. `delegated_at` is not validated at all. With a valid key and a signature generated over the supplied value, I reproduced successful verification for `null`, `false`, `0`, `[]`, and `{}`.

All of those calls returned normally at head `a96074921a2a2f967ce8fa9d7e11dd2aead889be`. The manifest schema gate may reject them in the enclosing verifier, but `verify_delegation_chain()` is an exported public API and is the boundary covered by #356 and this PR's new regression matrix.

Please validate `hop` as a JSON integer excluding booleans before the sequential comparison, reusing `_is_json_integer()` so the intentional `0.0` equivalence remains intact. Also establish that `delegated_at` is a string before it enters the signed preimage. Add re-signed controls showing these malformed JSON primitives fail with `ValueError` while valid timestamp strings remain accepted.

altrudev:
@carloshvp Addressed the blocking primitive-type gap on exact head `0256ffd8f68de1551fa674acbb14f133bada7e87`.

- `hop` now goes through the existing `_is_json_integer()` guard before the sequential comparison, so booleans are refused while zero-fraction JSON integers such as `0.0` remain accepted.
- `delegated_at` must be a string before it can enter the signed preimage.
- The `hop: false` regression deliberately keeps the valid signature over integer `0`, so the new structural guard—not signature failure—is load-bearing.
- The `delegated_at` matrix (`null`, `false`, `0`, `[]`, `{}`) re-signs each exact malformed value before verification, again pinning the structural boundary rather than crypto failure.
- Explicit valid timestamp-string and `hop: 0.0` controls remain accepted.

Exact-head verification is now complete: full CI passed and CodeQL passed on `0256ffd8f68de1551fa674acbb14f133bada7e87`. No other scope or preimage semantics were changed.

carloshvp (approved):
Re-reviewed exact head `0256ffd8f68de1551fa674acbb14f133bada7e87`. The blocking primitive-boundary issue is resolved.

The `hop` guard now rejects booleans before Python's bool/int equality can affect the sequential check, while preserving the intended RFC 8785 / JSON-Schema equivalence for `0.0`. The `delegated_at` guard runs before signed-preimage construction, and the new matrix re-signs each malformed value so the tests cannot pass through an incidental signature failure.

Local verification on this head:

- focused delegation tests: `108 passed`
- full Python suite: `1288 passed, 6 skipped`

All substantive GitHub checks are also green. Approving.

imran-siddique (approved):
Merging. This is the delegation-chain instance of the rule you have been applying all week, and the per-hop scoping is what makes it right.

Validating the root before root/manifest binding and root-depth reads, and validating each later hop **when verification reaches that hop** rather than all upfront, is the correct placement. Validating everything at the start would look tidier and would be wrong: it would do work on hops that a failure earlier in the chain means nobody should ever interpret, and it would report structural problems in a hop whose predecessor already invalidated the chain.

Keeping `_validate_hop_structure()` as the single verifier-side structural authority, and widening it to establish the object and primitive shapes downstream code already assumed, is better than adding checks at the sites that assumed them. One authority that can be read and audited beats five guards that agree today.

Closes #356. Fourteen checks, only the maintainer hold red.

That is six of yours merged across two days. The pattern running through all of them, establish the thing before interpreting under it, is a real seam and you have been working it systematically rather than opportunistically.

altrudev:
Hi Imran — quick Fellowship question.

The original Fellowship timeline indicated a September start, and since we’re now into September I wanted to check whether final selections have already been communicated or whether that process is still underway.

I haven’t seen a formal selection or non-selection notice, so I just want to make sure I haven’t missed anything.

Thanks.

PR #359, opened 2026-08-30
fix(transparency): fail closed on malformed Rekor responses

### LINKED ISSUE #358: verify_transparency_log_entry can raise on malformed Rekor 200 responses
## Current behavior

`verify_transparency_log_entry()` is a boolean verifier: existing tests establish that network errors, 404/500 responses, hash mismatches, and a missing Rekor body all return `False`.

At current `main` (`eb747f5fd610a8d7fa360e52faaea5db578e6b34`), however, only the HTTP GET itself is inside the exception boundary:

```python
try:
    response = httpx.get(...)
except Exception:
    return False

if response.status_code != 200:
    return False

body = response.json()
entry_data = next(iter(body.values()), {})
decoded = json.loads(base64.b64decode(entry_data.get("body", "e30=")))
actual_hash = decoded.get("spec", {}).get("data", {}).get("hash", {}).get("value", "")
```

A successful HTTP response is still remote, untrusted verification evidence. Several malformed but possible 200 responses escape the documented/established boolean boundary:

- invalid JSON -> response parser exception;
- JSON array/string/null -> `.values()` failure;
- entry value not an object -> `.get()` failure;
- invalid base64 body -> decode failure;
- decoded body not JSON -> `json.loads()` failure;
- decoded JSON scalar/list or malformed nested `spec`/`data`/`hash` containers -> chained `.get()` failure.

## Existing contract evidence

`python/tests/test_transparency.py` already pins `False` for:

- network exception;
- 404 and 500 responses;
- content-hash mismatch;
- missing `body` in an otherwise successful response.

So classifying an unparsable/malformed 200 response as `False` extends the existing verifier failure boundary rather than choosing a new trust policy.

## Expected behavior

Once the HTTP request succeeds, any response that cannot establish the expected Rekor entry/hash structure should return `False` just like missing/mismatching evidence. It should not terminate the caller with `JSONDecodeError`, `AttributeError`, `TypeError`, or a base64 decoding exception.

A valid matching Rekor response must remain `True`.

## Suggested narrow fix

Keep the network/status behavior unchanged and put response parsing + nested shape/hash extraction behind the same fail-closed boolean boundary. No new response schema or normalization layer is required for this issue.

## Regression matrix

- valid matching 200 -> `True`;
- invalid response JSON -> `False`;
- top-level array/string/null -> `False`;
- entry data scalar/list -> `False`;
- invalid base64 body -> `False`;
- base64-decoded non-JSON -> `False`;
- decoded JSON scalar/list -> `False`;
- malformed nested `spec` / `data` / `hash` -> `False`;
- existing missing-body, hash-mismatch, network, 404/500 controls unchanged.

No manifest schema, Rekor wire format, signed preimage, or trust-policy change is proposed. This is verifier exception-safety at an existing external evidence boundary.

altrudev:
## What

Closes #358.

`verify_transparency_log_entry()` now keeps malformed successful Rekor responses inside its existing boolean verification boundary instead of allowing response-parsing or nested-shape exceptions to escape.

The valid path and all existing status/network behavior are unchanged:

- matching 200 response -> `True`;
- non-200/network failure -> `False`;
- missing/mismatching evidence -> `False`;
- malformed/unparseable 200 response -> `False`.

## Why

The function already treats Rekor as external verification evidence and returns `False` for network errors, 404/500 responses, hash mismatches, and a missing body. But response parsing after a 200 was outside that failure boundary. Invalid JSON, non-object response shapes, invalid encoded bodies, or malformed nested hash containers could therefore leak `ValueError`, `AttributeError`, or `TypeError`.

A successful HTTP status does not establish the structure or meaning of the evidence body.

## Implementation

After the existing status check, response parsing now:

- establishes the top-level response and entry data as objects;
- establishes the encoded body as string/bytes;
- decodes/parses it fail-closed;
- establishes `spec`, `data`, and `hash` as objects before reading the expected hash value;
- returns `False` for malformed parsing/shape failures.

No new Rekor response schema or normalization layer is introduced.

## Regression matrix

A separate test file covers:

- valid matching 200 control;
- invalid response JSON;
- top-level array/string/null/integer/boolean;
- scalar/list/null entry data;
- invalid base64 and base64-decoded non-JSON;
- decoded JSON scalar/list/null values;
- malformed nested `spec`, `data`, and `hash` containers.

Existing tests already cover network errors, non-200 responses, missing body, and hash mismatch.

## Scope / non-claims

No manifest schema, Rekor wire format, signed preimage, cryptographic algorithm, or transparency trust policy changes. This is exception-safety at the existing external-evidence verification boundary.

AI-assistance disclosure: ChatGPT assisted with source triage, adversarial matrix design, implementation drafting, and DDC review. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/python/src/agent_manifest/_transparency.py
+++ b/python/src/agent_manifest/_transparency.py
@@ -175,7 +175,7 @@ def verify_transparency_log_entry(
         body = response.json()
         if not isinstance(body, dict):
             return False
-        entry_data = next(iter(body.values()), {})
+        entry_data: Any = next(iter(body.values()), {})
         if not isinstance(entry_data, dict):
             return False
         encoded_body = entry_data.get("body", "e30=")
--- a/python/src/agent_manifest/_transparency.py
+++ b/python/src/agent_manifest/_transparency.py
@@ -171,12 +171,32 @@ def verify_transparency_log_entry(
     if response.status_code != 200:
         return False
 
-    body = response.json()
-    entry_data: dict[str, Any] = next(iter(body.values()), {})
-    decoded = json.loads(base64.b64decode(entry_data.get("body", "e30=")))
-    actual_hash = (
-        decoded.get("spec", {}).get("data", {}).get("hash", {}).get("value", "")
-    )
+    try:
+        body = response.json()
+        if not isinstance(body, dict):
+            return False
+        entry_data = next(iter(body.values()), {})
+        if not isinstance(entry_data, dict):
+            return False
+        encoded_body = entry_data.get("body", "e30=")
+        if not isinstance(encoded_body, (str, bytes)):
+            return False
+        decoded = json.loads(base64.b64decode(encoded_body))
+        if not isinstance(decoded, dict):
+            return False
+        spec = decoded.get("spec", {})
+        if not isinstance(spec, dict):
+            return False
+        data = spec.get("data", {})
+        if not isinstance(data, dict):
+            return False
+        hash_data = data.get("hash", {})
+        if not isinstance(hash_data, dict):
+            return False
+        actual_hash = hash_data.get("value", "")
+    except (AttributeError, TypeError, ValueError):
+        return False
+
     return bool(actual_hash == expected_hash)
 
 
--- a/python/tests/test_transparency_malformed_response.py
+++ b/python/tests/test_transparency_malformed_response.py
@@ -1,0 +1,118 @@
+"""Regression vectors for malformed Rekor verification responses (#358)."""
+
+from __future__ import annotations
+
+import base64
+import hashlib
+import json
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from agent_manifest._canonicalize import canonicalize
+from agent_manifest._transparency import (
+    TransparencyLogEntry,
+    verify_transparency_log_entry,
+)
+
+
+ENTRY = TransparencyLogEntry(
+    log_id="test-log",
+    entry_id="test-entry",
+    inclusion_proof="proof",
+)
+MANIFEST: dict = {}
+
+
+def _response_with_json(value) -> MagicMock:
+    response = MagicMock()
+    response.status_code = 200
+    response.json.return_value = value
+    return response
+
+
+def _encoded(value) -> str:
+    return base64.b64encode(json.dumps(value).encode()).decode()
+
+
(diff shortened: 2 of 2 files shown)


imran-siddique:
Batch response for this cluster is here: https://github.com/agentrust-io/agent-manifest/pull/357#issuecomment-5472317549

Short version: the finding class is real and welcome. Your CI had never run, held under first-time-contributor gating, until I released 36 runs across your PRs an hour ago, and five of your eight are now red. Please fix those, sequence trace-spec#258 against #252 which touch the same two files, and tell me the order you want them reviewed in.

imran-siddique (approved):
Merging. Third of the three you queued, and the `mypy` failure from yesterday is cleared.

**The boundary was almost right, which is why the gap survived.** `verify_transparency_log_entry()` already returns `False` for network errors, non-200 responses, hash mismatches and a missing body. Someone reading it would reasonably conclude the function fails closed. Response parsing after a 200 sat just outside that boundary, so invalid JSON, a non-object shape, an invalid encoded body or a malformed nested hash container escaped as `ValueError`, `AttributeError` or `TypeError` from a function whose contract is a boolean.

An almost-complete failure boundary is worse than an obviously incomplete one, because it stops anyone looking.

**This sentence is the one I want in the commit message:**

> A successful HTTP status does not establish the structure or meaning of the evidence body.

Rekor is a third party. A 200 means the request was served. It says nothing about whether what came back is the shape we assumed, and this is the transparency path, so the caller is asking whether an independent anchor exists at all. Treating a well-served malformed answer as an exception rather than as "no usable evidence" puts the crash in the caller instead of the verdict in the result.

Keeping the valid path and every existing status behaviour unchanged, and moving only the parse failures inside the boundary, is the right scope. Nothing here changes what a good response means.

Thirteen checks, only the maintainer hold red. Closes #358.

That is your three. Holding #257, #258, #370, #596 and #597 as you proposed, and I will pick them up when you say they are ready rather than pre-empting your sequencing. cmcp#596 is currently red on `Lint`, if that is useful to know before you get to it.

Worth saying plainly: you took a batch response about volume and turned it into a queue with a peer review at the front, and all three merged in order without a round trip. That is the fastest way through this repository and it is now the arrangement I would like to keep.

Issue #343, opened 2026-08-26
Spec gap: Art. 50(1)/(3) disclosure is runtime enforced by AGT but produces no attestable evidence


FaresMallouli:
Section 9.1.2 (added in PR #277) correctly states the manifest has no field for Article 50(1)/(3) and that AGT's agent_os.transparency interceptor enforces it at the tool-call boundary and is the current answer. I traced that interceptor (agent-governance-toolkit, agent_os/transparency.py) and confirmed it's real: it blocks tool execution until confirm_disclosure or acknowledge_emotion_recognition is called for a session. But that enforcement produces no durable evidence, no hash, no signature, nothing outside the running process. An auditor asking whether a specific deployment actually delivered disclosure has no record to check, only source code to trust.

I'd like to propose an 11th manifest artifact, transparency_policy, that declares pre-deployment whether Art. 50(1)/(3) applies and binds a hash of the active interceptor configuration into the signed manifest, the same way policy_bundle binds Cedar policy today. This doesn't move disclosure delivery into the manifest, that correctly stays a runtime concern, but it makes the configuration tamper-evident rather than just trusted. Happy to scope this further as a fellowship proposal if there's interest.

imran-siddique:
Verified both halves, and the gap you name is real.

§9.1.2 says what you quote, and I checked the AGT side rather than taking it: `agent-governance-python/agent-os/src/agent_os/transparency.py` exists with its own test file, so the interceptor is genuinely the current answer rather than a plan.

**The timing is what makes this worth acting on.** §9.1.2 records Article 50 as applying from 2 August 2026 and as "the only EU AI Act obligation in force against agent deployments today". Three weeks in force. Everything else in section 9.1 is deferred to 2027 or 2028.

**And your framing is the one that matters.** The section already tells implementers they must not read the manifest as evidence of Article 50 compliance. What you are pointing out is that the runtime enforcement leaves nothing else to read either: an auditor asking whether a specific deployment delivered disclosure has source code to trust and no record to check. Enforcement that produces no evidence is the thing this project exists to fix, and finding it inside our own compliance mapping is the right kind of embarrassing.

## Status, so you are not guessing

[...]

PR #366, opened 2026-08-30
docs(spec): note Art. 50 runtime enforcement produces no attestable a…

FaresMallouli:
## What

Adds one paragraph to §9.1.2 stating that the runtime enforcement it names for Art. 50(1)/(3) produces no attestable artifact.

## Why

§9.1.2 tells implementers the manifest is not evidence of Article 50 compliance, but the runtime control it defers to leaves no durable record either — no hash, no signature, nothing outside the running process. An auditor has source code to trust and nothing to verify. This states that limitation rather than adding a new one. Per the direction in #343.

Scoped to Art. 50(1) and 50(3). The 50(2)/(4) marking gap has a different reason and is unchanged here.

## Spec impact

§9.1.2, non-normative. No new field, no schema change, no new obligation on producers. The Art. 50 table and the "Implementers MUST NOT" paragraph are unchanged. No conformance test IDs affected.

## Test plan

Documentation-only change to spec prose; no code paths touched.

## DCO

Refs #343

### DIFF
--- a/spec/agent-manifest-spec-v0.2.md
+++ b/spec/agent-manifest-spec-v0.2.md
@@ -1899,6 +1899,8 @@ Article 50 applies from **2 August 2026** and is the only EU AI Act obligation i
 | Art. 50(3) | Deployers of emotion recognition or biometric categorisation systems must inform exposed persons | **Not satisfied** in the manifest. Enforced at runtime by AGT's `agent_os.transparency` interceptor. Note that a manifest whose `hitl_record.approvals[].approved_scope` covers biometric tooling is evidence the deployment is in Art. 50(3) scope. |
 | Art. 50(4) | Deepfake and public-interest text disclosure | **Not satisfied.** Same gap as Art. 50(2). |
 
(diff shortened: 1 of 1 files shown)


imran-siddique (approved):
This is the half of #343 that needed no sponsor, and it is the right half to have written down.

§9.1.2 already told implementers not to read the manifest as evidence of Article 50 compliance. What it did not say is that the runtime enforcement leaves nothing else either, so a reader could reasonably infer the evidence existed somewhere outside the manifest. Naming the interceptor and stating plainly that it leaves no hash, no signature and no record outside the running process closes that inference.

"A party who cannot inspect that process has source code to trust" is the sentence that earns its place: it says what the reader is actually being asked to rely on.

Article 50 is the only EU AI Act obligation in force against agent deployments today, so of everything in section 9.1 this is the one where a reader is most likely to act on a wrong impression this year.

The manifest-side binding stays under `policy: take-internal` on #343. Nothing here weakens the case for it; if anything it states the gap the binding would close.

PR #367, opened 2026-08-30
docs(spec): require issuer key authorization

litzki-systems:
## What

Make issuer key authorization normative in section 5.3.3. Require signing key resolution and issuer authorization through an independent trust path before signature verification.

## Why

Close the issuer authorization gap tracked in issue #325. A key identifier identifies a signing key without establishing issuer authorization. Subject controlled identifiers must not establish the authorization path. Authorization failure and undetermined authorization remain distinct conditions, and neither permits a VALID result.

This also aligns the specification with the implementation gap described in issue #247, where the SDK enforces the authorization check more strictly than the specification currently requires.

## Spec impact

Update section 5.3.3 of `spec/agent-manifest-spec-v0.2.md`.

The change adds normative requirements for issuer key resolution and authorization. It adds no result codes, schema fields, or authorization mechanism. It applies the requirement across the v0.1 and v0.2 signature envelope semantics.

## Test plan

* [ ] Documentation only normative spec change
* [ ] No implementation changes
* [ ] No conformance vectors added in this PR

## DCO

### DIFF
--- a/spec/agent-manifest-spec-v0.2.md
+++ b/spec/agent-manifest-spec-v0.2.md
@@ -1441,10 +1441,6 @@ A `VALID` result means all of the following are true:
 - If `enforce_hitl` is `true`, at least one HITL approval is present, valid, not expired, and meets the `approval_method` requirement for the declared `risk_tier`
 - If `enforce_attestation` is `true`, `audit_key_sealed` is `true` in the attestation block
 
-**Issuer key authorization is not covered by a `VALID` result unless the verifier configures it (non-normative implementation note).** `signature.key_id` sits outside the signing pre-image (section 3.6), so it identifies a key without authorizing one. The reference SDK checks that the signing key is authorized for the declared `issuer` only when `VerificationContext.trusted_key_issuers` is populated; that field defaults to an empty map, and with it empty the check returns without evaluating anything. A deployment that does not populate it therefore obtains `VALID` results in which nothing has confirmed the signing key belongs to the claimed issuer.
-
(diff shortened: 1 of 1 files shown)


imran-siddique (approved):
This is the text I asked for, and it does the four things that mattered.

**Both requirements are normative and correctly ordered.** Resolution and authorization MUST complete *before* signature verification, which is the ordering that matters: a signature checked first and authorized second is a signature you have already spent effort trusting.

**The subject-controlled identifier prohibition is stated at the right level of abstraction.** This sentence is the one I would have written last and worst:

> This does not prohibit issuer and subject from being the same organization. It prohibits the authorization path from resolving through a subject-controlled identifier. The identity may be the same. The trust path may not.

That is the distinction people get wrong, and stating it as a permission rather than only as a prohibition is what stops a reader concluding self-issued manifests are banned.

**The stale note is gone.** §5.3's note saying the requirement was missing and tracked in #325 would have contradicted the section immediately below it. Removing it in the same PR was the right call and it is the part most likely to have been left behind.

**Authorization failure and undetermined authorization are distinct, and neither is `VALID`.** That is the discipline six separate threads have needed this week: absence of a determination is reported as absence rather than promoted to a verdict. Requiring the verifier to *distinguish* them, rather than only to refuse both, is what makes the result actionable for a relying party.

Envelope-agnostic phrasing covering v0.1 `signature.key_id` and v0.2 COSE `kid` was your point on the issue and it survived into the text intact.

All eleven substantive checks green: Lint, Type check, Security scan, AGT governance verify, Build check, and Test on 3.11/3.12/3.13 across ubuntu, windows and macos.

**One thing that is not yours.** §5.3.2 and now §5.3.3 both physically precede §5.3.1, and both use `#####` where §5.3.1 uses `####`. That ordering is already wrong on `main` (§5.3.2 sits above §5.3.1 there too), so you followed the existing convention rather than introducing anything. I am fixing it separately rather than asking you to.

Merging as the accepting maintainer under the arrangement on #325. Authorship yours; the requirement is mine to answer for. Thank you for waiting three days for a one-word answer and then turning it round in a day.

PR #369, opened 2026-08-30
feat: add memory checkpoint behavioral assessment harness

Knapp-Kevin:
## Summary

Implements the external, non-normative `MemoryCheckpointAssessment/0.1` reference harness discussed in #298.

The contribution adds deterministic behavioral evidence for candidate memory checkpoints without changing Agent Manifest verification or conformance semantics.

## What it adds

- typed assessment evidence models and a minimal retriever adapter boundary
- four retrieval invariants: correction precedence, anchor preservation, scope isolation, and state-conditioned differentiation
- full retriever-profile binding, including tie behavior and repeatability evidence
- state-reference stability checks so evidence cannot be bound to a checkpoint that changes during assessment
- a separate applicability gate for an explicitly adopted approval policy
- deterministic public reference vectors and a reproducible vector builder
- artifact-shape, portability, hardening, applicability, and canonicalization-boundary tests
- architecture documentation and MkDocs navigation

## Decision boundary

- assessment evidence records what was evaluated and what happened
- applicability determines whether that evidence can satisfy an adopted gate
- `pass` satisfies this gate only; it does not authorize deployment
- an applicable `fail` blocks the approval action while remaining valid evidence
- an applicable `indeterminate` remains a distinct “not established” result and does not satisfy the gate
- absence of applicable assessment evidence leaves the gate unsatisfied rather than implicitly passing
- historical evidence remains intact if an approval control is bypassed

Scope isolation is intentionally a negative confidentiality property. An empty retrieval contains no forbidden material and therefore satisfies that probe, but does not establish retrieval liveness; suites requiring liveness must include a positive invariant.

## Explicit limits

- no `spec/` or conformance change
- no checkpoint approval API
- no TRACE/runtime integration
- no new signing or envelope format
- no completeness or anti-selective-disclosure claim
- no full RFC 8785 portability claim while the remaining #322 U+2028 escaping defect is open

## Validation

Validated head: `321e3b3137b2c455b92db311ecc12fd9d6195f78`

- Ruff: passed
- strict mypy: passed
- Bandit: passed
- pip-audit: passed
- build/twine: passed
- AGT governance verify: passed
- Python 3.11, 3.12, 3.13 Ubuntu: passed
- Python 3.11 macOS: passed
- Python 3.11 Windows: passed
- CodeQL: passed

Python 3.11 Ubuntu: 1,312 collected, 1,305 passed, 6 skipped, 1 deliberate strict xfail for the remaining #322 canonicalization boundary. Total package coverage is 89.82%; the assessment harness is at 93% and the applicability gate at 90%.

Addresses #298.

### DIFF
--- a/docs/memory-checkpoint-assessment.md
+++ b/docs/memory-checkpoint-assessment.md
@@ -1,0 +1,186 @@
+# Memory checkpoint assessment
+
+`MemoryCheckpointAssessment/0.1` is a non-normative reference assessment for the retrieval behavior of a candidate memory checkpoint. It is tracked in [issue #298](https://github.com/agentrust-io/agent-manifest/issues/298).
+
+The problem is deliberately narrow. A checkpoint can be cryptographically valid, append-only, fresh, and within its update budget while still changing what the retriever selects in a harmful way. This assessment adds evidence about that retrieval behavior without changing the meaning of Agent Manifest verification.
+
+## What the assessment answers
+
+The assessment asks whether a candidate checkpoint preserves a declared set of retrieval invariants under a pinned retriever configuration.
+
+It does **not** decide whether stored content is true, whether an application response is correct, whether a memory write was authorized, or whether a checkpoint is safe in every possible sense. It also does not approve a checkpoint. The deterministic result depends only on the declared probes and retrieval evidence, not on a probabilistic judge.
+
+## Keep evidence and approval separate
+
+The assessment artifact records facts about a run. The approval gate decides whether those facts satisfy an adopted policy.
+
+That separation is important because a failed assessment can still be perfectly valid evidence. The checkpoint can remain cryptographically sound and its consistency proof can still verify. What changes is whether an approval action is allowed to proceed.
+
+The reference implementation keeps five concerns distinct:
+
+1. **Evidence validity**: is the assessment well formed and correctly bound to what it claims?
+2. **Behavioral outcome**: did the required probes return `pass`, `fail`, or `indeterminate`?
+3. **Applicability**: does this assessment match the checkpoint, probe suite, retriever profile, retrieval state, and evidence window required by the adopted gate?
+4. **Gate result**: does this one assessment gate allow the approval process to continue?
+5. **Audit history**: what actually happened, including an approval that should have been blocked?
+
+## Approval decision tree
+
+The gate is intentionally pass-only:
+
+```text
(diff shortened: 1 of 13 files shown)


imran-siddique (approved):
Merging. Saying plainly what I verified and what I did not, because 3,247 lines deserve that rather than an implied full read.

## What I verified

**The external boundary holds.** Only two files under `python/src/` are touched and both are new. Nothing in the existing verification path is modified, so "does not change Agent Manifest verification or conformance semantics" is checkable rather than asserted. That is what makes this mergeable without the sponsor question arising at all.

**The three-state gate is real.** `GateFailureReason` carries `no_applicable_assessment`, `applicable_assessment_failed` and `applicable_assessment_indeterminate` as distinct values, and the validator requires each to be reported explicitly rather than allowing a gate to fail without saying which. That is the ruling from #298 implemented rather than paraphrased:

- `no_applicable_assessment` is the omission bypass you identified. Making "nobody ran the assessment" a named outcome, instead of the silence it currently produces, is the whole point and it is the one most implementations would have left out.
- `applicable_assessment_indeterminate` keeps "not established" distinct from "established as failing". Five other threads in this codebase need that same distinction and most of them do not yet have it.

**The nav entry is present.** `mkdocs.yml` is in the diff. A docs page shipping without it is a recurring class here and you closed it without being asked.

**The ratio.** 964 lines of implementation against roughly 1,838 of tests plus 207 of reference vectors and a 247-line reproducible vector builder. Deterministic public vectors with a builder, rather than fixtures nobody can regenerate, is the part that will still be true in six months.

## What I did not verify

The 745-line assessment implementation, the four retrieval invariants (correction precedence, anchor preservation, scope isolation, state-conditioned differentiation), and the reference vectors themselves. I am merging on the boundary, the gate semantics, the test ratio and thirteen green checks, not on a line read of the invariants.

If any of the four invariants turns out to be wrong, it is wrong in a non-normative external harness that nothing else imports, which is the reason I am comfortable with that trade here and would not be for a change to `_verify.py`.

## One note for later

The `MemoryCheckpointAssessment/0.1` version marker is doing real work, and the vector README plus the builder mean a future revision can be shown to change or preserve outcomes rather than argued about. Worth keeping that discipline when 0.2 arrives.

Thank you for the five-concept restatement on #298 that preceded this. It is the reason the gate has three failure reasons instead of a boolean.

PR #370, opened 2026-08-30
fix(verify): negotiate version before current schema

### LINKED ISSUE #365: verify_manifest validates current schema before applying version negotiation
## Current behavior

The v0.2 specification makes version negotiation an ordering requirement:

> A verifying party MUST inspect the `version` field before processing any other field. If the verifying party does not support the declared version, it MUST return `INCOMPATIBLE_VERSION` and MUST NOT return `VALID` or `MISMATCH`.

`verify_manifest()` repeats the same contract in its docstring: an unsupported or missing version returns `INCOMPATIBLE_VERSION` without further processing.

At current `main` (`eb747f5fd610a8d7fa360e52faaea5db578e6b34`), the implementation does the opposite ordering for dict manifests:

```python
manifest_id = manifest.get("manifest_id", "unknown")
...
schema_violations = _strict_schema_violations(manifest)
if schema_violations:
    result.result = OverallResult.MISMATCH
    ...
    return result

version = manifest.get("version")
if version not in SUPPORTED_MANIFEST_VERSIONS:
    result.result = OverallResult.INCOMPATIBLE_VERSION
    return result
```

`_strict_schema_violations()` runs the current Pydantic `Manifest` model with `extra="forbid"` before the unsupported version is classified.

That means an unsupported-version document that is shaped like today's model can reach `INCOMPATIBLE_VERSION`, but a future-version document that uses a field or structure unknown to the current model can be returned as `MISMATCH` first.

For example, take the existing `AM-VEC-007` unsupported-version control (`version: "0.3"`, expected `INCOMPATIBLE_VERSION`) and add any plausible future-only top-level member. The current schema treats that member as an extra-field violation and returns `MISMATCH` before the version gate runs.

## Why this matters

A future field is exactly what version negotiation exists to protect against. Applying today's schema first causes the verifier to make a semantic assertion about a format it has already declared it does not understand.

`MISMATCH` means the evidence was processed under a known contract and contradicted it. `INCOMPATIBLE_VERSION` means the verifier has no authority to interpret that format. The two states answer different questions.

This is also observable across implementations: an implementation that follows §2.4 literally will stop on `version`, while the reference SDK can descend into current-schema validation first.

## Existing vector gap

`AM-VEC-007` says:

> "Unsupported manifest version is rejected before verifying."

but its manifest deliberately remains close enough to the current structure that the schema compatibility exceptions let it through to the version check. It therefore proves the result for one old-shape/future-version case, not the required ordering against a future-shaped document.

## Expected invariant

For a dict manifest:

1. establish only enough top-level structure to read `version` safely;
2. if the version is absent/unsupported, return `INCOMPATIBLE_VERSION` without interpreting current-version schema semantics;
3. only for a supported version, apply that version's schema and subsequent verification rules.

The missing-version treatment should follow the existing normative wording/project decision; this issue does not attempt to redefine it.

## Regression matrix

- current supported version + valid current shape -> existing behavior;
- current supported version + malformed current field -> `MISMATCH` as today;
- unsupported version + otherwise current-compatible shape -> `INCOMPATIBLE_VERSION` (existing AM-VEC-007 control);
- unsupported version + unknown future top-level field -> `INCOMPATIBLE_VERSION`, not `MISMATCH`;
- unsupported version + future-shaped nested artifact -> `INCOMPATIBLE_VERSION`, not a current-schema defect;
- non-object input should retain a documented/fail-closed public API boundary rather than creating a new host-language exception.

No schema rule for a supported version is being relaxed. This is verifier prerequisite ordering: **version authority precedes schema interpretation**.

altrudev:
## What

Closes #365.

For dict manifests, `verify_manifest()` now performs version negotiation before applying the current Pydantic schema or interpreting any current-version field other than the `version` discriminator.

An unsupported or missing version returns `INCOMPATIBLE_VERSION` immediately with `manifest_id="unknown"`. Supported versions continue through the existing schema, revocation, validity, signature, artifact, delegation, HITL, attestation, and transparency pipeline unchanged.

## Why

Spec §2.4 requires implementations to inspect `version` before processing other fields. The existing ordering ran `_strict_schema_violations()` first, so a future manifest carrying fields unknown to today's closed-world schema could be classified as `MISMATCH` before the verifier reached its version gate.

The ordering matters because current schema is not authoritative over a format the verifier has already declared it does not understand.

## Regression evidence

A focused regression matrix now establishes that:

- a future-only field does not invoke current schema;
- malformed current-version shapes under an unsupported version still return `INCOMPATIBLE_VERSION`;
- missing version follows the same incompatible-version gate;
- the unsupported-version path reads only the `version` discriminator;
- supported `0.1` manifests still apply the current schema and reject an unknown field.

The existing language-neutral `AM-VEC-007` is strengthened rather than adding a duplicate vector. It now carries a signed future-only artifact member that today's schema rejects, so its expected `INCOMPATIBLE_VERSION` result is load-bearing on the required ordering.

## Scope

No schema rule, wire format, signature algorithm, current-version validation rule, result vocabulary, or supported-version set changes. This is prerequisite ordering only.

The branch is one commit directly on current upstream `ad01af774d198aafb3b9191040d88cfbc515f4ec`, one commit ahead / zero behind, and changes exactly four files: the verifier, focused regression test, vector generator, and `AM-VEC-007`.

Upstream moved once after the original review freeze: #367 changed only `spec/agent-manifest-spec-v0.2.md`. That file does not overlap any #365 file. The candidate was therefore reconstructed on the new upstream tree using the exact four previously reviewed blobs, then the one-commit/four-file diff gate was repeated.

## Review notes

Relevant gates:

- Need ≠ Authority: PASS. §2.4 supplies the normative ordering authority.
- Semantic propagation: PASS. The change stops at version prerequisite ordering; supported-version semantics remain in the existing pipeline.
- Non-Redundant Build: PASS. Existing `AM-VEC-007` was strengthened instead of adding a duplicate conformance rule.
- Failure-path independence: PASS for the static design. The portable vector and focused read-guard exercise the ordering through different mechanisms.
- Transition lineage: PASS. Current predecessor `ad01af77…` is pinned; the only intervening upstream change was independently checked as spec-only/non-overlapping; the reconstructed candidate is one commit ahead / zero behind with the same four reviewed blobs.
- Executable gate: pending this PR's repository CI and maintainer review at the time of this description; no manual workflow rerun is claimed here.

AI-assistance disclosure: ChatGPT assisted with source triage, falsification design, regression construction, rebase lineage verification, and drafting. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/python/src/agent_manifest/_verify.py
+++ b/python/src/agent_manifest/_verify.py
@@ -631,6 +631,16 @@ def verify_manifest(
     if isinstance(manifest, (bytes, bytearray)):
         return _verify_cose_envelope(bytes(manifest), context, revocation_store)
 
+    # Version negotiation is the prerequisite for every current-schema
+    # interpretation below (spec 2.4). Read only the discriminator first. A
+    # future-shaped manifest must not be classified against today's schema.
+    version = manifest.get("version")
+    if version not in SUPPORTED_MANIFEST_VERSIONS:
+        return VerificationResult(
+            manifest_id="unknown",
+            result=OverallResult.INCOMPATIBLE_VERSION,
+        )
+
     manifest_id = manifest.get("manifest_id", "unknown")
     result = VerificationResult(manifest_id=manifest_id, result=OverallResult.VALID)
     agent_uid = manifest.get("agent_id")
@@ -649,11 +659,11 @@ def verify_manifest(
     transparency_unverifiable = False
 
     # --- Schema validation (fail-closed). verify_manifest accepts a raw dict,
-    # so it must run the manifest through the Pydantic guards before trusting
-    # any field. This makes extra="forbid" (unknown fields), enum/type
-    # constraints, the expiry window, and timestamp parsing actually apply on
-    # the verify path. A malformed expires_at is a schema failure here, not a
-    # silently non-expiring manifest.
+    # so it must run a supported manifest through the Pydantic guards before
+    # trusting current-version fields. This makes extra="forbid" (unknown
+    # fields), enum/type constraints, the expiry window, and timestamp parsing
+    # actually apply on the verify path. A malformed expires_at is a schema
+    # failure here, not a silently non-expiring manifest.
     #
     # Required top-level claims fail closed. Missing runtime observations in the
     # VerificationContext still produce NOT_BOUND; legacy omissions nested in
@@ -668,13 +678,6 @@ def verify_manifest(
                 actual_hash=f"<{msg}>",
             ))
         result.mismatch_details = mismatches
-        return result
-
-    # --- Version negotiation (spec 2.2 / 2.4) - MUST be checked before
-    # verifying so unsupported manifests are never silently misinterpreted.
-    version = manifest.get("version")
-    if version not in SUPPORTED_MANIFEST_VERSIONS:
-        result.result = OverallResult.INCOMPATIBLE_VERSION
         return result
 
     # --- Revocation check (must happen before VALID can be returned)
--- a/python/tests/test_version_negotiation_order.py
+++ b/python/tests/test_version_negotiation_order.py
@@ -1,0 +1,103 @@
+from __future__ import annotations
+
+from typing import Any
+
+import agent_manifest._verify as verify_module
+from agent_manifest._verify import (
+    OverallResult,
+    RevocationStore,
+    VerificationContext,
+    VerificationResult,
+    verify_manifest,
+)
+
+
+FUTURE_VERSION = "9.9"
+VALID_MANIFEST_ID = "018f4a3b-2c1d-7e5f-a8b9-0d1e2f3a4b5c"
+
+
+def _verify(manifest: dict[str, Any]) -> VerificationResult:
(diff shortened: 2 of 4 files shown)


imran-siddique (approved):
Merging. The ordering is the substance here.

Applying the current Pydantic schema before establishing which version a manifest declares means an unsupported version fails somewhere inside a schema that does not apply to it, and the error a caller sees describes a field rather than the incompatibility. Returning `INCOMPATIBLE_VERSION` immediately, with `manifest_id="unknown"` because nothing has authorised reading that field yet, is both the correct behaviour and the honest one: a manifest we cannot interpret is not a manifest whose id we know.

Reading only the `version` discriminator before negotiation, and nothing else, is the part that makes this a real boundary rather than a reordering. Any other field read first would be interpreted under an assumed version.

Shipping `AM-VEC-007.json` with the change means the rule is pinned by an artifact rather than by a test body. Closes #365. Thirteen checks, only the maintainer hold red.

PR #372, opened 2026-08-31
docs(examples): fix broken links to renamed/nonexistent spec and tutorial files

rajnisht7:
## What

This PR fixes a dead link in `examples/README.md` that points to a spec file which doesnt exist and another of tutorial

## Why

`examples/README.md` links to `spec/agent-manifest-spec-v0.1.md`, but that file isn't in the repo only `spec/agent-manifest-spec-v0.2.md` is.

`examples/revocation/README.md` linked to `docs/tutorials/revocation.md`; but the actual file is `docs/tutorials/revocation-and-key-rotation.md` (the link text already said "Revocation and key rotation" just the path was stale).

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -32,4 +32,4 @@ pip install -e "python/"
 
 - [Getting started](../docs/getting-started.md)  -  step-by-step walkthrough of creating and verifying manifests with the Python SDK and CLI
 - [Tutorials](../docs/tutorials/)  -  detailed guides for delegation chains, revocation, HITL, hardware attestation, and server-side verification
-- [Spec](../spec/agent-manifest-spec-v0.1.md)  -  full data model and field cardinality rules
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
Checked both rather than taking them, against `main`:

```
spec/agent-manifest-spec-v0.1.md               404
spec/agent-manifest-spec-v0.2.md               200
docs/tutorials/revocation.md                   404
docs/tutorials/revocation-and-key-rotation.md  200
```

Both dead, both now pointing at what exists. The second one is the more interesting of the two: the link text already read "Revocation and key rotation", so the page was renamed and only the path was left behind. That is the version of this bug that survives a proofread, because the sentence is correct and only the target is wrong.

Two lines, eleven checks, only the maintainer hold red. Merging.

PR #373, opened 2026-09-01
fix(attestation): don't let the Azure platform label stand in for manifest binding proof

### LINKED ISSUE #363: verify_attestation_chain routes unsupported platform labels through the SNP verifier
## Current behavior

`verify_attestation_chain()` documents platform-specific hardware-signature dispatch. At current `main` (`eb747f5fd610a8d7fa360e52faaea5db578e6b34`) the dispatch is:

```python
platform = getattr(report, "platform", "") or ""
if platform == "intel-tdx":
    signature = _verify_tdx_signature_step(...)
elif platform in ("tpm", "aws-nitro"):
    signature = _verify_tpm_signature_step(...)
else:
    signature = _verify_snp_signature_step(...)
```

The `else` is therefore not "SEV-SNP"; it is **every other string**, including typos, unsupported future values, `opaque`, and an empty/unknown label.

That matters because the backend does not independently re-check the report's platform label. Given otherwise valid SNP report bytes, VCEK material, matching `report_data`, and no rejecting measurement policy, changing only `AttestationReport.platform` from `"amd-sev-snp"` to an unsupported value still selects the SNP verification path.

The existing full-pass SNP test provides a direct mutation control: its cryptographic evidence does not depend on the Python `platform` string, so the label and the verification backend are currently separable.

## Why this is security-relevant

The result object reports one overall `passed` boolean and signature status. A caller can therefore receive a successful SNP cryptographic appraisal attached to a report whose declared platform says something else.

This is a platform-dispatch/binding problem, not a claim that the SNP signature itself can be forged.

It is also inconsistent with the repository's own provider boundary:

- `SEVSNPProvider` emits `platform="amd-sev-snp"`;
- `TDXProvider` emits `platform="intel-tdx"`;
- Azure has its own `azure-cvm-sev-snp` path because its manifest binding is vTPM-rooted rather than direct guest-controlled SNP `REPORT_DATA`;
- `OPAQUEProvider` is documented as not implemented and fail-closed at construction.

An unsupported platform label should not silently inherit the SNP verifier merely by falling through the dispatch.

## Expected invariant

The selected verification procedure must be authorized by the declared platform/profile. Unknown or unsupported platform values must fail closed rather than be interpreted as a known backend.

I am intentionally not choosing whether the result should use `FAILED`, `NOT_IMPLEMENTED`, or a distinct unsupported-platform status. That result vocabulary is a maintainer decision.

## Suggested narrow direction

Use explicit platform-to-verifier dispatch for every supported platform and an explicit fail-closed branch for everything else. Keep Azure's distinct binding model distinct from direct SEV-SNP rather than folding it into the same default.

Also update the `AttestationReport.platform` field comment, whose current illustrative values (`"tpm" | "sev-snp" | "tdx" | "opaque"`) no longer match the concrete labels emitted by the providers.

## Regression matrix

- valid `amd-sev-snp` + valid SNP evidence -> existing result;
- same exact evidence with `platform="opaque"` -> must not pass as SNP;
- same with unknown/empty/future platform label -> fail closed;
- valid `intel-tdx` -> existing TDX path unchanged;
- valid TPM/AWS Nitro -> existing TPM path unchanged;
- Azure label remains governed by its vTPM-rooted verification semantics rather than direct SNP fallback.

No hardware signature algorithm, certificate chain, manifest hash pre-image, or platform trust root is changed by this issue. The correction is explicit backend selection and platform/evidence binding.

rajnisht7:
## What

Fixes Azure attestation so a correctly signed SNP report with a wrong (or missing) manifest binding cannot pass, and closes the unsupported-platform dispatch gap from #363.

Closes #363

## Why

Azure CVM does not put the manifest hash in REPORT_DATA (the paravisor binds the vTPM AK there). The real binding is a vTPM quote over a PCR derived from the manifest hash, chained through runtime data into the signed SNP report.

An earlier approach treated the `azure-cvm-sev-snp` platform label (or a caller-supplied flag) as proof that this binding was checked. That was wrong: a valid SNP signature with a wrong PCR, or with no quote/AK evidence at all, could still return `passed=True`.

## Changed:
- `verify_attestation_chain` establishes the full Azure binding itself from evidence on the report (PCR value in the AK-signed quote, PCR bank/index, AK signature, full AK identity `(n, e)` vs HCLAkPub, runtime data -> REPORT_DATA). No caller boolean.
- Platform dispatch is an explicit allow-list; unknown labels get `NOT_IMPLEMENTED` instead of falling through to the SNP verifier.
- Fail-closed on missing, malformed, or mismatched evidence (including bad runtime-data JSON shapes).

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO


carloshvp (changes requested):
Requesting changes for one blocking trust-boundary problem: the Azure platform label is being treated as proof that the manifest binding was verified.

This branch sets `report_data_matched = True` whenever `platform == "azure-cvm-sev-snp"`, but `verify_attestation_chain()` neither calls `AzureCVMProvider.verify_manifest_in_report()` nor receives an authenticated result from it. The reason string says the check happens separately, but the public function has no ordering or precondition that enforces that statement.

I reproduced this on the submitted head `f8f1113ae3684f587d520b08e76016fca1811c3c` with one self-consistent, correctly signed SNP report:

- the report carried a deliberately wrong PCR value;
- `AzureCVMProvider.verify_manifest_in_report(report, manifest)` returned `False`;
- this changed `verify_attestation_chain()` returned `passed=True`, `report_data_matched=True`, and `signature=VERIFIED` for the same report;
- the same signed report also returned `passed=True` for three unrelated `expected_manifest_hash` values (`11…`, `22…`, and `ff…`).

The new positive test encodes the same gap: its report has no `pcr_read`, no demonstrated vTPM/PCR verification, and still asserts `passed is True` solely because the platform string is Azure.

`report_data` being inapplicable is not equivalent to the manifest binding matching. Conflating those states turns the absence of this check into positive verification.

Before this can merge:

1. Do not synthesize `report_data_matched=True` from the platform label.
2. If this public result is meant to pass for Azure, establish the Azure composite binding inside the verification boundary: the SNP/VCEK chain, SNP runtime-data binding to the vTPM AK, the AK-signed TPM quote/PCR evidence, and the PCR value derived from the expected manifest hash.
3. If that composite verifier is outside this PR's scope, represent Azure report-data binding as not applicable without allowing `passed=True`; keep the result fail-closed until separately supplied evidence is actually consumed and verified.
4. Replace the current positive test with load-bearing controls: missing PCR evidence, a mismatching PCR, and arbitrary expected manifest hashes must fail; only authenticated matching Azure binding may pass.

The focused test file is green (`14 passed`), but it is green around the wrong security property. The direct-SNP non-Azure control is useful and should remain.

Also account for #363 when rebasing: Azure may share the SNP signature primitive, but it remains a distinct binding profile and must be reached through explicit platform dispatch.

imran-siddique:
Maintainer confirmation on both halves, because the premise and the remedy point different ways here.

**@rajnisht7, the problem you found is real and it is ours.** On Azure CVM, SEV-SNP runs under a Hyper-V paravisor and the attestation is vTPM-rooted: `REPORT_DATA` binds the vTPM AK, not the manifest hash, and the guest does not control that field. That is documented in this module and in `LIMITATIONS.md` because we established it on live SNP hardware, not because it was assumed. A function requiring `REPORT_DATA == manifest_hash` on every platform genuinely cannot pass a correct Azure report, and that is worth fixing.

**@carloshvp is right that this fix cannot land, and the reproduction is the reason.** Setting `report_data_matched = True` on the strength of the platform label makes the label the evidence. The function neither calls `AzureCVMProvider.verify_manifest_in_report()` nor receives an authenticated result from it, so the reason string saying the check happens separately is a promise the code does not keep:

> the report carried a deliberately wrong PCR value; `verify_manifest_in_report(report, manifest)` returned `False`; this changed `verify_attestation_chain()` returned `passed=True`, `report_data_matched=True`, and `signature=VERIFIED` for the same report

A correctly signed report with a wrong PCR returning `passed=True` is the exact outcome the function exists to prevent. And it is invisible to CI: fourteen checks are green, because nothing in the suite asserts that the two agree.

**The shape of a fix that would land.** The Azure path needs the binding *established*, not assumed:

- `verify_attestation_chain()` takes the manifest-binding result as an input it can check, or calls the provider itself, rather than inferring it from the platform string.
- `report_data_matched` stays `False` unless something authenticated set it. If the caller did not supply the binding result, the honest outcome is **not established**, distinct from both verified and mismatched, and it must not report `passed=True`. That is the same three-state discipline we have been applying across the other verifiers this week: absence of a check is never a pass.
- A test in @carloshvp's shape, a correctly signed report with a wrong PCR, asserting the function does **not** return `passed=True`. That case is the one that would have caught this, and it does not currently exist.

The general rule I want on the record, since this is the second time this month a platform label has stood in for a verification: a `platform` value says which verification procedure applies. It is never evidence that the procedure ran.

carloshvp (changes requested):
Re-reviewed exact head `f7577a7a5d481aeb6f185bebc07fc0f254d1859d`. The explicit platform dispatch and the Azure `None` / `False` fail-closed paths are fixed, but the positive Azure path still has one blocking authentication gap.

`azure_manifest_binding_verified=True` is described as an authenticated result from `AzureCVMProvider.verify_manifest_in_report()`. That provider method does not authenticate the Azure PCR evidence: it reads the plain `report.raw["pcr_read"]` string and compares it with the expected extension. It does not verify `quote_msg`, `quote_sig`, `quote_pcrs`, or the AK public key, and it does not establish that this AK is the one bound into the signed SNP runtime data.

I reproduced this on the submitted head with a cryptographically valid synthetic SNP report and VCEK/ASK/ARK chain. Its `raw` data contained a fabricated matching `pcr_read` and deliberately omitted all of `ak_pub_pem`, `quote_msg`, `quote_sig`, `quote_pcrs`, and runtime-data binding evidence. The result was:

```text
provider_binding_result=True
has_ak_evidence=False
signature=verified
report_data_matched=True
passed=True
```

The returned reason also says the result was authenticated, although no AK evidence existed. Moving the decision into a caller-supplied boolean therefore does not establish the composite chain or bind that decision to this report and manifest.

Before this can pass review, the Azure positive path must consume a result from a boundary that actually verifies the composite evidence for this exact report and manifest: PCR value inside an AK-signed TPM quote, the quote signature under the AK, and the AK/runtime-data binding inside the signed SNP report. A freely supplied boolean, or a boolean derived only from `pcr_read`, is not sufficient. Please add negative controls that omit or tamper the AK quote/signature/PCR/runtime-data links and prove none can produce `passed=True`.

The existing tests are green locally (`70 passed, 2 skipped` focused; `1322 passed, 6 skipped, 1 xfailed` full), but the current positive test sets the boolean directly and therefore does not exercise the claimed authentication boundary.

carloshvp (changes requested):
Re-reviewed exact head `709a993f580a31ce0650d5ad10e9bbf7bf7ded80`. Removing the caller-supplied boolean and verifying the composite evidence inside `verify_attestation_chain()` is the right direction. Three blocking gaps remain in the new verifier.

1. The signed TPM quote's PCR selection is discarded. `verify_azure_manifest_binding()` checks only `quote.pcr_digest` and assumes it represents one SHA-256 PCR at the manifest index. I changed the signed selection bitmap from `000001` (PCR 16) to `000002` (PCR 17), kept the same digest, re-signed the exact quote with the runtime-data-bound AK, and the full chain still returned `signature=verified`, `report_data_matched=True`, `passed=True`. The verifier must parse and require the expected bank and PCR index, not merely a digest that could have come from another PCR. Since `AzureCVMProvider` supports a configurable PCR index, that expected index needs to enter this boundary as verifier configuration rather than being inferred from self-reported `raw` data.

2. The claimed exact AK identity check compares only the RSA modulus. I changed the runtime-data JWK exponent to `e=3` (`"Aw"`) while keeping the PEM key at `e=65537`, rebound that runtime data into a newly signed SNP report, and the chain still returned `passed=True`. An RSA public key is `(n, e)`; both values must match the `HCLAkPub` JWK.

3. The fail-closed exception contract is not complete. Runtime data containing valid JSON `{"keys": 1}`, correctly hash-bound into the signed SNP report, makes `ak_modulus_hex_from_runtime_data()` iterate an integer and leaks `TypeError: 'int' object is not iterable` through `verify_attestation_chain()`. The helper says malformed input returns `None` and the composite verifier says it never raises, so nested JSON types and top-level argument types need explicit guards or equivalent exception handling.

Local results on this head:

```text
wrong_pcr_selection: selected_bitmap=000002, signature=verified, binding=True, passed=True
different_exponent: pem_e=65537, runtime_e=3, signature=verified, binding=True, passed=True
malformed_keys_shape: raised TypeError ('int' object is not iterable)
```

The existing focused tests pass (`81 passed, 2 skipped`) and the full suite passes (`1333 passed, 6 skipped, 1 xfailed`); mypy, ruff, and `git diff --check` are clean. Please add load-bearing negative controls for all three cases.

One merge-order note: this branch extracts `verify_ak_signature()` with the old signature-prefix discriminator that the already-approved #375 replaces with key-shape classification. Rebase or resolve the overlap so #375's fix is preserved rather than reintroduced through the helper extraction.

carloshvp (changes requested):
Re-reviewed exact head `2a1d395d53ea669c6f88faa5dfc4684a0273d235`. The three findings from the prior review are fixed: the signed PCR bank/index is checked from verifier configuration, the complete RSA public key `(n, e)` is compared, malformed runtime-data container shapes fail closed, and the #375 key-shape signature framing is preserved.

One blocking malformed-evidence path remains. `ak_public_numbers_from_runtime_data()` and `ak_modulus_hex_from_runtime_data()` decode JWK members with permissive `base64.urlsafe_b64decode()`. That decoder silently discards characters outside the alphabet. I prefixed the runtime-data `HCLAkPub.n` value with `!!!!`, rebound that exact JSON into a newly signed synthetic SNP report, and kept every other link valid. The current verifier returned:

```text
signature=verified
report_data_matched=True
passed=True
```

The runtime data is cryptographically committed, but its JWK modulus is not valid base64url. This contradicts the helpers' malformed-input contract and lets malformed evidence receive positive verification credit.

Please enforce the unpadded URL-safe alphabet before decoding both `n` and `e` (and in the modulus-only helper), then use strict decoding. Add full-chain negative controls for illegal prefixes/suffixes and standard-base64 `+`/`/` aliases, demonstrating that a correctly rebound and signed report still cannot pass with malformed JWK encoding.

The focused attestation/provider/TPM suites otherwise pass (`100 passed, 2 skipped`), the full suite passes (`1352 passed, 6 skipped, 1 xfailed`), and substantive CI is green. The green suites do not currently cover this encoding alias.

carloshvp:
Re-reviewed current head c5c912f5ff4cc7b2babb6c3ee55c2424237a8471.

The previous blocking findings are resolved. Azure can reach passed=True only after verifying the complete composite binding: the configured SHA-256 PCR selection and expected value inside the signed TPM quote, the AK signature, the full AK identity (n, e) in runtime data, and the runtime-data binding into the signed SNP report. Missing, mismatched, malformed, wrong-type, or merely self-reported evidence fails closed.

The regression suite now includes load-bearing negative controls for the platform-label bypass, arbitrary manifest hashes, wrong PCR value/selection/bank, AK exponent mismatch, malformed base64url, malformed runtime-data shapes, missing evidence, signature tampering, and runtime-data/SNP binding failures.

Local verification: 1537 passed, 6 skipped, 1 xfailed.

Approving from my side.

carloshvp (dismissed):
Re-reviewed exact head c5c912f5ff4cc7b2babb6c3ee55c2424237a8471. The prior blocking findings are resolved, including authenticated Azure composite binding, configured PCR selection, complete AK identity, strict base64url validation, malformed-evidence fail-closed behavior, and load-bearing negative controls. Local verification reported 1537 passed, 6 skipped, 1 xfailed. Approving.

rajnisht7:
@carloshvp, thank you for the review, just pushed a commit fixing SNP fallback and empty JWK validation, can you  have a look at it again

carloshvp (dismissed):
Re-reviewed updated exact head 910ba66430881dc41a5a35ae7fed00ef00efa81d. The follow-up correctly prevents explicit empty SNP evidence from being replaced by report.quote and rejects empty JWK modulus/exponent values, with regression coverage. Isolated validation: 107 passed, 1 skipped across attestation-chain, SNP, and provider tests. Substantive CI is green. Approving.

imran-siddique (commented):
Review/test follow-up on `dc3f761f37d3`.

The Azure path now checks the quote PCR bank/index and digest, AK signature, both RSA public numbers, and runtime-data binding to the SNP report; the caller still requires the hardware signature chain. The platform label and supplied booleans cannot substitute for those checks. The full local suite and fresh hosted CI passed. The previous reviews were dismissed after later pushes, so this head needs fresh eligible review after its sign-offs are corrected.

Full local `python -m pytest -q` (Windows, Python 3.12, existing review environment): **1595 passed, 6 skipped, 1 xfailed, 2 warnings in 44.51s**.

Contribution requirement still outstanding: these commits have no `Signed-off-by:` trailer: `52271a8c`, `b2f6b4b7`, `22393a4e`, `dc3f761f`. Please have the respective commit authors add their own DCO sign-offs, including merge commits, then request review of the resulting head.

Fresh workflow evidence: [34048199970](https://github.com/agentrust-io/agent-manifest/actions/runs/34048199970) (success).

imran-siddique (dismissed):
Re-reviewed `585efc7339ecc0a139bf2731745e14c21c54d05d`. The sign-off rewrite is tree-identical to the head reviewed and tested earlier, all commits now carry DCO trailers, and fresh CI passes.

The Azure binding verifies the configured PCR bank/index and digest, quote signature, complete AK public key, and runtime-data/SNP binding; the hardware signature remains mandatory. Full local suite: 1595 passed, 6 skipped, 1 xfailed.

imran-siddique:
Merged current main into `d2dc1dce45b9e90cb1a16199d296ad81718e48b0` after #377 landed. The only conflicted file was CHANGELOG.md; both security entries are retained. I also checked the automatically combined CRL code: authenticated loading and strict signature decoding are both preserved.

Full local suite on the combined tree: 1609 passed, 6 skipped, 1 xfailed. The merge commit carries my DCO sign-off. Fresh CI must finish on this head, and because I pushed this repair, I am requesting an independent review from @pforest before protected merge.

imran-siddique (dismissed):
Reviewed the merge with main at d2dc1dce45b9e90cb1a16199d296ad81718e48b0. Both changelog entries are preserved. The combined CRL implementation retains authenticated-load failures from #377 and strict signature decoding from this PR. Full local suite: 1609 passed, 6 skipped, 1 xfailed; fresh hosted CI and CodeQL pass. The merge commit is signed off.

I pushed the conflict repair, so this approval does not satisfy independent last-push approval. Review requested from pforest; protected auto-merge must wait for that eligible review.

imran-siddique:
The head is unchanged from the completed review, but GitHub now reports a merge conflict after the latest main updates. Please rebase and preserve the attestation-binding regressions. The resulting head will need CI and an independent last-push review before the existing protected auto-merge can proceed.

imran-siddique:
Bookkeeping only, so the next read starts from facts rather than from the review history. No verdict on the diff here.

**The branch was rewritten after my 6 September review.** That review was of the merge with main at `d2dc1dce`. The head is now a single commit, `f143d406`, and it has diverged from `d2dc1dce` rather than building on it. That is why both of my reviews and all four of @carloshvp's read as dismissed: nothing was withdrawn, the commits they were attached to no longer exist. Anyone reading the review timeline should treat every prior review as applying to a tree that is gone.

**What I verified on the new head, because it is what the merge repair existed to protect.** #377 merged on 6 September, and its behaviour survives the rewrite: `_revocation.py` at `f143d406` still carries authenticated mode via `trusted_signer_key`, the `UntrustedCRL` failure for a malformed authenticated load, and the `CRL-001` rule that a null or empty `revocation_signature` raises `InvalidSignature` rather than `ValueError`. Both changelog entries are present, including #377's `--crl-trusted-key` entry. So the squash did not silently drop the other PR's work, which was the specific risk.

**Current state.** Two commits behind main, and main has moved further since: #375, #376, #378, #404, #406 and #408 have all landed today, including the `prometheus-client` lockfile fix that was failing `test_docs_first_manifest` on every PR in this repository. Only `Policy: Awaiting maintainer review` is red, which is the hold gate, and one of its two rollup entries is stale.

**What it is waiting on.** An independent review. I requested @pforest on 6 September and it has not arrived. My own approval cannot satisfy the last-push rule on the tree I pushed to, and in any case that tree is no longer the head, so the question is now simply that 2,377 lines of Azure attestation verification have had no review of the current commit from anyone.

rajnisht7:
CI is failing on `fuzz_cose`, it is because `attach_receipt`/`attach_attestation`/`attach_approvals` can crash on re-encode if the unprotected header contains malformed CBOR value (cbor2's internal break-marker sentinel), as it wasn't validated before being copied back out. 

Opened a PR for the same #411

carloshvp (approved):
Reviewed exact head `a145e464d242c60b866a3d848adf107183136f80` after the branch rewrite and current-main integration.

The security boundary remains intact. Azure reaches `passed=True` only after the verifier authenticates the complete chain for this exact report and manifest: configured SHA-256 PCR selection and value inside the AK-signed TPM quote, AK signature, full AK identity `(n, e)` in runtime data, runtime-data binding into the signed SNP report, and the required VCEK certificate chain. Platform labels, caller-supplied booleans, unauthenticated `pcr_read`, missing evidence, malformed types, noncanonical base64url, mismatched PCR bank/index/value, AK mismatch, signature tampering, and runtime-data/SNP mismatch all fail closed. The current tree also preserves the merged TPM signature key-shape classification and authenticated CRL behavior.

Independent isolated validation on this head:

- Full suite: `1677 passed, 6 skipped`.
- Focused attestation, provider, TPM, signing, TRACE, and type suites: `368 passed, 2 skipped`.
- Repository CI lint rule: clean.
- mypy: no issues in 27 source files.
- `git diff --check`: clean.
- Merge simulation against current `main` (`d5262c88`) is conflict-free; both commits have DCO sign-offs.

The remaining `Fuzz changed code` failure is the repository-wide malformed-CBOR issue tracked and fixed separately by #411, not a regression in this attestation diff. `Policy: Awaiting maintainer review` is the review gate this approval addresses. Approving.

imran-siddique:
State, re-measured today so my own read starts from facts rather than from the review history. No verdict on the diff.

**The defect is still live on `main`.** `_attestation.py`'s dispatch there is `if intel-tdx / elif tpm, aws-nitro / else: SNP`, so any unrecognised platform label still falls through to the SEV-SNP verifier. That is #363 as filed. And `_azure_verify.py` still does not exist on `main`, so an Azure CVM report is checked today against `REPORT_DATA`, which on Azure binds the vTPM AK rather than the manifest hash. Both halves of what this fixes are real right now.

**Clean against current `main`.** Trial-merged just now, after #417 and the other merges since you last rebased: no conflicted files, `CHANGELOG.md` included. Nothing is owed on your side for that.

**Green.** All 14 checks pass on `71eff9e9`, `Fuzz changed code` among them, which confirms your 9 September diagnosis: that red was #411's cbor2 break-marker defect on `main` and never this branch.

**Why it has not merged, which is not what the review timeline suggests.** @carloshvp approved on 10 September and `reviewDecision` still reads `REVIEW_REQUIRED`, so the approval is not the gate being missed. CODEOWNERS routes `/python/` to `@agentrust-io/sdk-maintainers`, and that rule overrides the `*` rule carloshvp's review sits under. 17 of this PR's 18 files are under `python/`, so their approval covers `CHANGELOG.md` and nothing else. The PR needs an approval from pforest, Qiang-Xu or me.

That is mine to resolve and I am saying so rather than leaving it looking like a queue. Nothing is waiting on you.

**What I still owe before I approve it.** I have not read all 2385 lines. The 61 tests in `test_attestation_chain.py` pass and the refusals are named for the holes rather than asserted generically, a valid SNP signature with the wrong PCR among them. The two areas I want a second pass on are `_azure_verify.py`'s runtime-data key parsing and the strict base64url decoding, which is where several of those tests are aimed and where a permissive decoder would quietly widen what counts as evidence.

imran-siddique (approved):
Reviewed 71eff9e97342761f58987ea88626aede25947cec.

Reviewed the Azure evidence path through runtime-data AK parsing, quote signature and PCR binding, and the SNP verification chain. Unsupported platform labels are rejected explicitly. Malformed evidence, strict base64url inputs, PCR selection, and RSA exponent mismatch have regression coverage. A clean trial merge with main at 8070bbb6e064e6609a36682e8d82ae68bbf0ed4f includes the catalog fix in #418. Combined local suite: 1,775 passed, 10 skipped; required Ruff checks and mypy on 27 source files pass. Disabling the exponent comparison makes the targeted mismatch regression fail; restoring it passes. These are fixture-based checks, not live Azure hardware validation. No blocking findings.

PR #374, opened 2026-09-01
fix(tdx): bind verification to signed quote header

### LINKED ISSUE #364: verify_tdx_quote does not bind verification to the quote's declared TDX header
## Current behavior

`verify_tdx_quote()` is documented as fully verifying an **Intel TDX v4 DCAP quote**. The module defines the production header constants:

```python
_TDX_QUOTE_VERSION = 4
_ATT_KEY_TYPE_ECDSA_P256 = 2
_TEE_TYPE_TDX = 0x81
```

and `parse_tdx_quote(..., strict=True)` checks the version and `tee_type`.

At current `main` (`eb747f5fd610a8d7fa360e52faaea5db578e6b34`), however, the actual full verifier never calls that strict parser:

```python
def verify_tdx_quote(...):
    parsed = parse_tdx_quote_signature(quote)
    ...
```

`parse_tdx_quote_signature()` slices the signed header/body and parses the certification material but does not validate `version`, `tee_type`, or `att_key_type`.

There is a second gap in the supposedly strict parser itself: `parse_tdx_quote()` unpacks `att_key_type` but currently checks only `version` and `tee_type`; `_ATT_KEY_TYPE_ECDSA_P256` is never enforced there either.

So the verifier can apply the P-256 TDX verification procedure to a self-consistent signed quote whose **signed header declares a different version, a non-TDX TEE type, or a different attestation-key type**.

## Why this is a binding problem

The header is inside `signed_body`, so this is not a claim that an attacker can simply flip those bytes on an existing quote without breaking the signature.

The narrower problem is that the signed declaration and the verifier's interpretation are not constrained to agree. A valid signature establishes authenticity of the declared header; it does not make an unsupported header mean TDX v4/P-256.

The repository already contains the right primitive (`parse_tdx_quote(strict=True)`); the full verifier simply does not consume that prerequisite, and that primitive itself omits the `att_key_type` check.

## Existing test gap

`test_parse_rejects_wrong_tee_type` exercises `parse_tdx_quote()`, not `verify_tdx_quote()`.

The full-chain test always builds:

```python
version = 4
att_key_type = 2
tee_type = 0x81
```

so it never proves that the **verifier** refuses a self-consistent quote declaring another header. A useful mutation control is to construct/re-sign the synthetic quote with one header field changed while keeping the same test trust chain; today the cryptographic steps are still driven as P-256 TDX.

## Expected invariant

A successful TDX verification must establish both:

1. the signatures/chain are valid; and
2. the signed header declares the exact quote version, TEE type, and attestation-key type the verifier implementation is interpreting.

Signature validity must not silently authorize a different declared profile to be parsed under TDX-v4 semantics.

## Suggested narrow direction

Make the full verifier establish the strict quote header before any signature/certification semantics are accepted, and include `_ATT_KEY_TYPE_ECDSA_P256` in that strict header check.

No new algorithm or wire format is needed.

## Regression matrix

Using the existing synthetic full-chain builder, produce self-consistent signed controls for:

- version 4 / key type 2 / TEE type 0x81 -> existing True path;
- wrong quote version -> reject;
- wrong `tee_type` -> reject as not TDX;
- wrong `att_key_type` -> reject rather than verify under P-256 anyway;
- ordinary body/signature tampering -> existing behavior unchanged;
- wrong pinned root -> existing chain failure unchanged.

No exploitability claim beyond profile-confusion is made. This is verification-profile binding: the procedure applied must match the signed header that declares what the evidence is.

altrudev:
Closes #364.

## What

`verify_tdx_quote()` now establishes the signed production TDX profile before accepting any signature or certification semantics:

- `version == 4`;
- `att_key_type == 2` (`ECDSA_P256`);
- `tee_type == 0x81` (`TDX`).

The full verifier calls `parse_tdx_quote(..., strict=True)` as its profile prerequisite. The strict parser now checks the previously omitted attestation-key type as well as version and TEE type.

`strict=False` remains available for diagnostic field extraction only; its documentation now states that it does not authorize cryptographic interpretation of a declared profile.

## Regression evidence

The existing synthetic full-chain builder is parameterized at the signed header. Each unsupported-header control is therefore self-consistent and re-signed by the same generated attestation key rather than byte-flipped after signing:

- production header `(4, 2, 0x81)` keeps the existing successful path;
- version `5` is refused by the full verifier;
- attestation-key type `3` is refused by the full verifier;
- TEE type `0x00` is refused by the full verifier;
- the strict parser directly refuses key type `3`;
- `strict=False` remains a diagnostic parse control.

Existing body-tamper, pinned-root, certificate-validity, QE-binding, signature-section, and chain behavior is left unchanged.

## Scope

No algorithm, wire format, certificate-chain rule, trust root, report-data binding, or signature preimage is changed. This patch only binds the implemented TDX-v4/P-256 verifier to the signed header that declares that profile.

AI-assistance disclosure: ChatGPT assisted with source triage, mutation-oriented regression design, implementation drafting, and exact-diff review. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/python/src/agent_manifest/_tdx_verify.py
+++ b/python/src/agent_manifest/_tdx_verify.py
@@ -195,13 +195,12 @@ def parse_tdx_quote(quote: bytes, *, strict: bool = True) -> TdxQuote:
 
     Args:
         quote: the raw DCAP quote bytes.
-        strict: when ``True`` (default), enforce the production layout —
-            ``version == 4`` and ``tee_type == 0x81`` — raising on anything
-            else. Pass ``strict=False`` to parse the header/body of an
-            otherwise well-formed quote whose version/tee_type differ (e.g.
-            synthetic test vectors), extracting the fields without asserting the
-            production TDX identity. Signature verification
-            (:func:`verify_tdx_quote`) is unaffected and always strict.
+        strict: when ``True`` (default), enforce the production header —
+            ``version == 4``, ``att_key_type == 2`` (ECDSA-P256), and
+            ``tee_type == 0x81`` — raising on anything else. Pass
+            ``strict=False`` only for diagnostic field extraction from an
+            otherwise well-formed quote; that mode does not authorize any
+            cryptographic interpretation of the declared profile.
     """
     if len(quote) < _QUOTE_HEADER_LEN + _TD_REPORT_LEN:
         raise TdxVerificationError(
@@ -212,6 +211,10 @@ def parse_tdx_quote(quote: bytes, *, strict: bool = True) -> TdxQuote:
     if strict:
         if version != _TDX_QUOTE_VERSION:
             raise TdxVerificationError(f"unsupported TDX quote version {version} (expected 4)")
+        if att_key_type != _ATT_KEY_TYPE_ECDSA_P256:
+            raise TdxVerificationError(
+                f"unsupported TDX attestation key type {att_key_type} (expected 2)"
+            )
         if tee_type != _TEE_TYPE_TDX:
             raise TdxVerificationError(f"not a TDX quote: tee_type {tee_type:#x}")
     body = quote[_QUOTE_HEADER_LEN:_QUOTE_HEADER_LEN + _TD_REPORT_LEN]
@@ -248,11 +251,12 @@ def verify_tdx_quote(
 ) -> bool:
     """Fully verify an Intel TDX v4 DCAP quote (all four steps, fail-closed).
 
-    Returns True only when the attestation-key signature, the QE binding, the
+    Returns True only when the signed quote header declares the production
+    TDX-v4/ECDSA-P256 profile, the attestation-key signature, the QE binding, the
     PCK signature over the QE report, and the PCK chain up to the pinned Intel
     SGX Root CA all check out. Raises :class:`TdxVerificationError` on a
-    malformed quote / broken chain or if ``cryptography`` is unavailable; returns
-    False on a well-formed-but-invalid signature.
+    malformed/unsupported quote or broken chain, or if ``cryptography`` is
+    unavailable; returns False on a well-formed-but-invalid signature.
 
     Every certificate in the PCK chain must be within its validity period (see
     :func:`._cert_chain.check_validity_period`); an expired PCK leaf,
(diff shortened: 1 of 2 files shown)


altrudev:
@carloshvp Implemented the accepted #364 boundary at exact head `1f703a760527f247dae143e33054b03388582d30`.

The full verifier now establishes `parse_tdx_quote(..., strict=True)` before any signature/certification semantics, and the strict parser now requires version 4, ECDSA-P256 attestation key type 2, and TDX tee_type 0x81. `strict=False` is documented as diagnostic extraction only.

The existing synthetic full-chain builder is parameterized at the signed header, so the wrong-version, wrong-key-type, and wrong-TEE controls are re-signed self-consistent quotes rather than post-signature byte flips. The valid production-header path and the existing tamper/root/certificate/QE/PCK behavior remain unchanged.

Exact-head gates: CI SUCCESS; CodeQL SUCCESS. PR is mergeable and the diff is one commit / two files. The PR is still Draft only because the GitHub integration's Ready-for-review mutation currently errors on an upstream GraphQL `fullDatabaseId` field; no code/test gate is outstanding.

imran-siddique (approved):
Merging. The ordering is the whole point of this change and it is right.

Establishing `version == 4`, `att_key_type == ECDSA_P256` and `tee_type == 0x81` **before** accepting any signature or certification semantics is the correct sequence, because interpreting a signature under an unverified profile means the profile is deciding how to read the evidence while being part of the evidence. Adding the previously omitted attestation-key type to the strict parser closes the gap that mattered most: an accepted key type is what tells you which algorithm the signature is even in.

**The `strict=False` documentation change carries more weight than its size.** "Does not authorize cryptographic interpretation of a declared profile" is the same distinction I have just written on #373, where a `platform` label was standing in for a completed verification. A declared profile says which procedure applies. It is never evidence that the procedure ran, and a diagnostic parser whose output is used as authorization is how that confusion enters a codebase.

**The regression evidence is built correctly**, and this is the second time this week I have said that about your tests: parameterising the synthetic full-chain builder at the signed header, so each unsupported-header control is self-consistent and re-signed by the same generated key rather than byte-patched, means a rejection is attributable to the header rule. A byte-patched control fails for two reasons at once and proves neither.

Thirteen checks, only the maintainer hold red. Closes #364.

PR #375, opened 2026-09-02
fix(tpm): classify signature envelopes by key shape

noah-ing:
## What

Classify legacy bare TPM quote signatures from the attestation key's signature shape instead of the untrusted `sigAlg` prefix:

- RSA byte input exactly one modulus wide remains in the documented bare PKCS#1 v1.5 lane.
- EC byte input must decode as a complete DER ECDSA signature to remain in the legacy lane.
- All other byte input is parsed strictly as a marshalled `TPMT_SIGNATURE`.

The regression coverage includes the committed reference vectors, supported and unsupported two-byte envelope mutations, malformed EC framing, and a fixed public synthetic vector whose valid 2048-bit bare RSA signature begins with `0x0014`.

## Why

This is a narrow follow-up to the typed signature work in #320 and the reference vectors in #323.

The current discriminator treats a byte string as `TPMT_SIGNATURE` only when its first two bytes name a supported scheme. As a result:

1. Changing only a real envelope's `sigAlg` to an unsupported value moves it into the legacy bare-signature path, changing an explicit unsupported-algorithm error into a generic `False` result.
2. A valid modulus-sized bare RSA signature whose cryptographic bytes happen to begin with `0x0014`, `0x0016`, or `0x0018` is misclassified as an envelope and rejected.

The patch makes framing depend on the AK representation, not attacker-controlled envelope metadata. Invalid mutated evidence remains rejected; its diagnostic becomes explicit. Documented legacy bare RSA inputs with scheme-like random prefixes can now verify normally.

The fixed compatibility vector is synthetic public test cryptography. It did not come from a TPM and establishes no hardware provenance.

## Spec impact

None. This is an SDK-only parsing and legacy-compatibility correction; no normative text or conformance IDs change.

## Test plan

- [x] `pytest -v` equivalent full suite passes: 1,340 passed, 6 skipped, 1 xfailed
- [x] `mypy src/agent_manifest` passes: 26 source files
- [x] `ruff check src/ tests/ --select E,F,W --ignore E501` passes
- [x] New and updated tests cover the change: 78 focused TPM tests pass
- [x] Spec changelog requirement is not applicable; the Unreleased SDK changelog is updated

Additional verification:

- `bandit -r src/agent_manifest -c pyproject.toml` — 0 findings
- `pip-audit` — no known vulnerabilities
- fresh wheel and sdist build — passed
- Twine 7.0.0 metadata checks — both distributions passed
- isolated install/API/CLI smoke from both wheel and sdist — passed
- `detect-secrets` 1.5.0 on changed files — 0 findings
- fixture hashes and direct PKCS#1 v1.5/SHA-256 verification — passed
- `git diff --check` and `git fsck --full` — passed

## DCO


carloshvp (dismissed):
Reviewed exact head `cb67bf1802a9a5af9b9e4fbac8216a1b00f8a6bf`. The framing change is sound.

Classification now depends on the public-key type from the already validated AK certificate chain. For RSA, only a modulus-width byte string remains in the documented bare PKCS#1 v1.5 lane; ordinary `TPMT_SIGNATURE` envelopes are longer and therefore reach the strict parser, including unsupported-scheme diagnostics. For EC, only a complete DER ECDSA value remains bare, while malformed or marshalled input reaches strict envelope parsing. The supported scheme/hash mutation controls continue to fail cryptographically.

I also verified the committed compatibility fixture independently: all four documented SHA-256 values match the decoded artifacts, the signature is 256 bytes with the intended `0x0014` prefix, and the fixture directory contains no private key material.

Local verification on this head:

- focused TPM tests: `78 passed`
- full Python suite: `1340 passed, 6 skipped, 1 xfailed`
- mypy: `26 source files` clean
- ruff: clean
- `git diff --check`: clean

Approving.

imran-siddique (dismissed):
Reviewed the TPM signature framing change and fixed RSA collision fixture. All 63 TPM verification and reference-vector tests pass locally. Restoring the previous verifier produces four failures, including the valid bare-RSA signature with a scheme-like prefix and explicit rejection of an unsupported envelope scheme.

imran-siddique (approved):
Reviewed the TPM signature framing change and fixed RSA collision fixture. The original 63 focused tests passed, and restoring the prior verifier produced four failures. After updating with main's stricter certificate-chain verifier, the fixture's missing CA extension caused one failure among 82 combined tests. I re-certified the same AK public key under a synthetic CA with the required constraints, preserving the quote and 0x0014-prefixed signature. All 82 combined TPM and certificate-chain tests now pass locally, and the repaired head's code checks are green.

noah-ing:
Thanks, Imran — I appreciate the careful review and the certificate-fixture update to keep this regression compatible with the stricter chain checks. Thank you as well for following the reports through to fixes and publication, and for the credit.

imran-siddique (approved):
Re-reviewed 18c1c242fa21253324da21ae691776ff4ccbdc0a. This corrects the missing sign-off on our own earlier merge commit and replays our signed fixture repair. The contributor's original commit is unchanged. A git tree comparison against previously reviewed 563ae4e095493e831e659a3d10d6e60a034fb3b6 is identical, so the earlier 82-test TPM/certificate-chain evidence still applies to the file contents. Fresh hosted CI is green. The repair still requires independent last-push approval.

imran-siddique (commented):
Review/test follow-up on `18c1c242fa21`.

The discriminator uses the attestation key signature shape, preserving both bare signatures and TPMT_SIGNATURE envelopes. The synthetic certificate fixture now satisfies CA constraints; the sign-off repair preserved the reviewed tree. The full local suite and fresh hosted CI passed. No additional defect found in this diff. Since I pushed the repair, protected auto-merge still needs an eligible independent approval.

Full local `python -m pytest -q` (Windows, Python 3.12, existing review environment): **1441 passed, 6 skipped, 1 xfailed, 2 warnings in 20.83s**.

Fresh workflow evidence: [34047334663](https://github.com/agentrust-io/agent-manifest/actions/runs/34047334663) (success).

PR #377, opened 2026-09-02
fix(cli): verify CRL revocation records against a trusted signer key

rajnisht7:
## What

This PR adds --crl-trusted-key to manifest verify so that --crl-path revocation records can actually be cryptographically verified, instead of always loading in FileCRL's unauthenticated mode.

## Why

FileCRL already supports trusted_signer_key for verifying revocation record signatures but the CLI never exposed a flag for it manifest verify --crl-path constructed FileCRL with no signer key. This means anyone who can write to it or intercept the CRL file can delete a revocation record to un-revoke a compromised manifest, or fabricate one, and the CLI accepts it silently.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,36 @@
 
 ### Fixed
 
+- **[SECURITY][CLI]** `manifest verify` gained `--crl-trusted-key PATH` to
+  authenticate `--crl-path` revocation records against the revoking
+  authority's public key. Previously the CLI always constructed `FileCRL`
+  without `trusted_signer_key`, so every CLI-driven CRL load ran in
+  `FileCRL`'s unauthenticated mode regardless of intent: a party who could
+  write to or intercept the CRL file could fabricate a revocation record
+  for a legitimate manifest, and `manifest verify` would accept it without
+  complaint. `--crl-trusted-key` is now required to authenticate the CRL;
+  omitting it still works (unchanged default) but now prints a loud
+  warning, and passing it without `--crl-path` fails cleanly instead of
+  being silently ignored. Same pattern as the `--approver-key` gap fixed
+  in 0.11.1.
+
+- **[SECURITY][CRL]** `FileCRL._load()` now fails closed in authenticated
+  mode: a malformed, unsigned, tampered, or wrong-key record used to be
+  silently skipped, which meant a party able to corrupt one line of an
+  authenticated CRL file could make that manifest register as *not*
+  revoked — the exact outcome `--crl-trusted-key` exists to prevent. Any
+  such record now raises `CRLIntegrityError` (surfaced by the CLI as a
+  clean, non-zero-exit error) instead of silently disappearing from the
+  in-memory set before `is_revoked()` runs.
+
+- **[DOCS]** Corrected an overclaim in the `--crl-trusted-key` help text
+  and `manifest verify` docs: authenticating records that are present in
+  a CRL file is not the same as proving the file is complete. Per-record
+  signatures cannot detect that a valid line — or the entire file — was
+  deleted, so `--crl-trusted-key` does not by itself prevent un-revocation
+  by deletion. Closing that gap needs a signed, versioned CRL
+  snapshot/digest mechanism, which is not yet implemented.
+
 - **[SDK]** `_check_manifest_binding()` no longer masks a malformed
   `artifacts` or `artifacts.policy_bundle` as merely absent. Truthy
   non-objects (`"str"`, `[1]`, `True`) already correctly reported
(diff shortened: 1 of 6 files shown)


carloshvp (dismissed):
Reviewed exact head `8082d20c4550b9ef74c2f6f752728f3fd7ab7d38`. Requesting changes because authenticated CRL mode currently fails open.

When `trusted_signer_key` is supplied, `FileCRL._load()` catches every record parse or signature-verification failure and silently continues. The CLI then sees no cached record and classifies the manifest as not revoked. Starting with a genuinely signed revocation, I changed one signature character and loaded the file with the correct trusted key. The result was:

```text
tampered_record_count=0
is_revoked=False
```

Deleting the record produced the same result. The new wrong-key regression test explicitly expects `VALID`, so the test currently codifies authentication failure as evidence of non-revocation. A party able to tamper with a record can therefore achieve the same un-revocation outcome this PR says the trusted-key path prevents.

For authenticated mode, malformed, unsigned, tampered, or wrong-key records must make CRL loading or manifest verification fail closed through an observable error/result state. They must not disappear from the set before `is_revoked()` runs. Please add enclosing CLI tests proving those cases cannot produce `VALID`.

There is also a separate guarantee mismatch: per-record signatures authenticate records that remain in the file, but they cannot prove that the list is complete or detect deletion of a valid line. Either add an authenticated snapshot/completeness mechanism (for example, a signed CRL envelope with version/freshness and a digest over the complete set), or narrow the PR body, changelog, documentation, and warning to claim authenticity only for records that are present. The current claim that `--crl-trusted-key` prevents un-revocation by deletion is not established by this implementation.

The four focused CLI tests pass and substantive CI is green, but one of those tests encodes the fail-open behavior above.

rajnisht7:
@carloshvp Thank you for the review, have made the changes:

- `FileCRL._load()` now raises `CRLIntegrityError` on any malformed/unsigned/tampered/wrong-key record in authenticated mode instead of skipping it. CLI converts this into a clean non-zero exit.

- The completeness/deletion gap is real and separate narrowed the docs/warning/changelog to say `--crl-trusted-key` authenticates records that are present, not that the file is complete.

- Fixed the test that was wrongly asserting `VALID` for a wrong-key record

carloshvp:
The original security blocker is resolved on `daf4f4d5d03d0bba445ac850d71112a7195cfd0d`.

Authenticated CRL mode now fails closed on malformed, unsigned, tampered, or wrong-key records; one invalid record invalidates the complete load, and the CLI surfaces this as a non-zero verification error. The tests exercise these cases at both `FileCRL` and CLI boundaries.

The claims are also correctly narrowed: per-record signatures authenticate records that remain present but cannot detect deletion or prove CRL completeness. The explicit deletion test accurately pins that known limitation.

Local verification on this head: `1352 passed, 6 skipped, 1 xfailed`.

The branch is now materially behind current `main` and has diverged from it. Please update/rebase it and confirm the full suite remains green. Assuming no substantive conflict-resolution changes to the CRL path, I see no remaining security blocker.

rajnisht7:
@carloshvp thank you for the review, have made the changes:

- `FileCRL` docstring: now says invalid/absent signatures raise `CRLIntegrityError` in authenticated mode, not "skipped"; added a note that a missing file behaves the same as an emptied one
- Removed the duplicated completeness caveat from `--crl-trusted-key`'s help text and regenerated `docs/api-reference/cli.md`
- Added test for pinning authenticated mode missing file behavior

imran-siddique (commented):
Review/test follow-up on `a19126f1c464`.

The authenticated loader now rejects invalid present records as a whole instead of silently dropping revocations. The documentation consistently retains the completeness limitation: whole-record/file deletion remains undetectable with per-record signatures. The previous functional and documentation findings are addressed; the full local suite and fresh hosted CI passed. After the sign-offs below are fixed, request a fresh review to clear the outstanding changes-requested state.

Full local `python -m pytest -q` (Windows, Python 3.12, existing review environment): **1481 passed, 6 skipped, 1 xfailed, 2 warnings in 23.31s**.

Contribution requirement still outstanding: these commits have no `Signed-off-by:` trailer: `8c7c1c92`, `b06fdc0d`, `a19126f1`. Please have the respective commit authors add their own DCO sign-offs, including merge commits, then request review of the resulting head.

Fresh workflow evidence: [34048107189](https://github.com/agentrust-io/agent-manifest/actions/runs/34048107189) (success).

imran-siddique (approved):
Re-reviewed `8dbfc1b888b92dfcf8281e6c966eaa7eb2580f31`. The sign-off rewrite is tree-identical to the head reviewed and tested earlier, all commits now carry DCO trailers, and fresh CI passes.

Authenticated CRL loading fails on invalid present records and the CLI surfaces that failure. The docs and tests correctly retain the inability to detect whole-record deletion. Full local suite: 1481 passed, 6 skipped, 1 xfailed.

PR #378, opened 2026-09-03
fix(delegation): verify_hitl_approval leaks host-language exceptions on malformed input

### LINKED ISSUE #360: verify_hitl_approval leaks host-language exceptions on malformed approval shapes
## Current behavior

`verify_hitl_approval()` is exported from the package public API, and its docstring states two rejection boundaries:

- `InvalidSignature` when signature verification fails;
- `ValueError` when required fields are missing or the approval has expired.

At current `main` (`eb747f5fd610a8d7fa360e52faaea5db578e6b34`), structure is interpreted before those boundaries are established:

```python
duration = approval.get("approved_scope", {}).get(
    "approval_duration_seconds", 0
)
...
pre = _approval_pre_image(
    manifest_id=manifest_id,
    approved_at=approval["approved_at"],
    approved_scope=approval["approved_scope"],
    approver_id=approval["approver_id"],
)
sig = approval["approval_signature"]
pad = 4 - len(sig) % 4
sig_bytes = base64.urlsafe_b64decode(...)
```

Malformed external values can therefore escape through incidental Python exceptions rather than the documented public API contract. Examples:

- non-object `approval` -> `AttributeError` from `.get()`;
- truthy non-object `approved_scope` -> `AttributeError` from the nested `.get()`;
- missing `approved_at`, `approved_scope`, `approver_id`, or `approval_signature` -> `KeyError`;
- truthy non-numeric `approval_duration_seconds` -> `TypeError` in `timedelta(seconds=...)`;
- non-string `approval_signature` -> `TypeError` from `len()`;
- malformed base64url can escape the signature-verification boundary through the decoder.

The main `verify_manifest()` path catches several of these around its call to `verify_hitl_approval()`, so this is **not** a claim that every malformed HITL record crashes the top-level manifest verifier. The defect is narrower: the separately exported helper does not satisfy the exception contract it documents.

This is the adjacent boundary explicitly left out of #356 so the delegation-chain repair could remain independently reviewable.

## Expected behavior

A caller using the exported `verify_hitl_approval()` API should be able to rely on its documented rejection types. Malformed approval structure should be classified before expiry arithmetic, pre-image construction, or signature decoding consumes fields.

## Suggested narrow direction

Establish the object/field shapes this helper reads before interpreting them, then map malformed approval structure to `ValueError`. Keep cryptographic signature failure as `InvalidSignature`.

This does not require changing HITL authority semantics, approval expiry semantics, the signed pre-image, or the manifest schema.

## Regression matrix

At minimum:

- non-object approval;
- missing required signed fields;
- `approved_scope` = string/list/bool;
- malformed/non-numeric `approval_duration_seconds`;
- non-string and malformed-base64url signature values;
- valid unexpired approval control;
- validly signed expired approval still reports the existing expiry failure;
- wrong key/signature still reports `InvalidSignature`.

No claim of cryptographic bypass is made. This is public-verifier exception-safety and prerequisite ordering.

kingztech2019:
## What

`verify_hitl_approval()` now establishes the shape of its input before interpreting any field, so malformed input raises the documented `ValueError` instead of an incidental `AttributeError`/`KeyError`/`TypeError`.

## Why

Closes #360. The function's docstring promises exactly two rejection boundaries (`InvalidSignature`, `ValueError`), but malformed input reached expiry arithmetic, pre-image construction, and signature decoding before either boundary was established:

- non-object `approval` -> `AttributeError` from `.get()`
- truthy non-object `approved_scope` -> `AttributeError` from the nested `.get()`
- missing `approved_at`/`approved_scope`/`approver_id`/`approval_signature` -> `KeyError`
- truthy non-numeric `approval_duration_seconds` -> `TypeError` in `timedelta()`
- non-string `approval_signature` -> `TypeError` from `len()`

Malformed base64 in `approval_signature` already raised `ValueError` by inheritance (`binascii.Error` subclasses it) and now gets an explicit, clearer message too.

## Spec impact

None. SDK-only exception-safety fix; no change to HITL authority semantics, approval expiry semantics, the signed pre-image, or the manifest schema.

## Test plan

- [x] `pytest -v` passes (1422 passed, 6 skipped, 1 xfailed — no regressions)
- [x] `bandit -r src/agent_manifest` passes
- [x] New tests cover the full regression matrix from #360: non-object approval, each missing required field, non-object `approved_scope`, non-numeric duration, non-string signature, malformed base64, non-string `approved_at`/`approver_id` (9 parametrized tests, 53/53 passing in `test_delegation_hitl.py`)
- [x] N/A — SDK-only, no spec change

## DCO

### DIFF
--- a/python/src/agent_manifest/_delegation.py
+++ b/python/src/agent_manifest/_delegation.py
@@ -516,18 +516,59 @@ def verify_hitl_approval(
 
     Raises:
         InvalidSignature: If the approval signature is invalid.
-        ValueError: If required fields are missing or the approval has expired.
+        ValueError: If required fields are missing, malformed, or the
+            approval has expired.
     """
     import base64
-    from datetime import datetime, timezone
+
+    # Establish the shapes this function reads before interpreting them, so a
+    # malformed approval always produces the documented ValueError rather than
+    # an incidental AttributeError, KeyError, or TypeError from whichever line
+    # happens to touch the bad field first.
+    if not isinstance(approval, dict):
+        raise ValueError(f"HITL approval must be an object, got {type(approval).__name__}")
+
+    for field in ("approved_scope", "approved_at", "approver_id", "approval_signature"):
+        if field not in approval:
+            raise ValueError(f"HITL approval is missing required field '{field}'")
+
+    approved_scope = approval["approved_scope"]
+    if not isinstance(approved_scope, dict):
+        raise ValueError(
+            f"HITL approval.approved_scope must be an object, got "
+            f"{type(approved_scope).__name__}"
+        )
+
+    approved_at_str = approval["approved_at"]
+    if not isinstance(approved_at_str, str):
+        raise ValueError(
+            f"HITL approval.approved_at must be a string, got "
+            f"{type(approved_at_str).__name__}"
+        )
+
+    approver_id = approval["approver_id"]
+    if not isinstance(approver_id, str):
+        raise ValueError(
+            f"HITL approval.approver_id must be a string, got {type(approver_id).__name__}"
+        )
+
+    sig = approval["approval_signature"]
+    if not isinstance(sig, str):
+        raise ValueError(
+            f"HITL approval.approval_signature must be a string, got {type(sig).__name__}"
+        )
 
     # HITL-003: enforce approval expiry before verifying signature
-    duration = approval.get("approved_scope", {}).get("approval_duration_seconds", 0)
+    duration = approved_scope.get("approval_duration_seconds", 0)
     if duration:
-        approved_at_str = approval.get("approved_at", "")
+        if not isinstance(duration, (int, float)):
+            raise ValueError(
+                f"HITL approval.approved_scope.approval_duration_seconds must be "
+                f"numeric, got {type(duration).__name__}"
+            )
         try:
             approved_at = datetime.fromisoformat(approved_at_str.replace("Z", "+00:00"))
-        except (ValueError, AttributeError) as e:
+        except ValueError as e:
             raise ValueError(f"HITL approval has invalid approved_at: {e}") from e
         if datetime.now(timezone.utc) > approved_at + timedelta(seconds=duration):
             raise ValueError(
@@ -537,11 +578,15 @@ def verify_hitl_approval(
 
     pre = _approval_pre_image(
         manifest_id=manifest_id,
-        approved_at=approval["approved_at"],
-        approved_scope=approval["approved_scope"],
-        approver_id=approval["approver_id"],
+        approved_at=approved_at_str,
+        approved_scope=approved_scope,
+        approver_id=approver_id,
     )
-    sig = approval["approval_signature"]
-    pad = 4 - len(sig) % 4
-    sig_bytes = base64.urlsafe_b64decode(sig + ("=" * pad if pad != 4 else ""))
+    try:
+        pad = 4 - len(sig) % 4
+        sig_bytes = base64.urlsafe_b64decode(sig + ("=" * pad if pad != 4 else ""))
+    except ValueError as e:
+        raise ValueError(
+            f"HITL approval.approval_signature is not valid base64url: {e}"
+        ) from e
     Ed25519Verifier(approver_public_key)._pub.verify(sig_bytes, pre)
--- a/python/src/agent_manifest/_delegation.py
+++ b/python/src/agent_manifest/_delegation.py
@@ -561,7 +561,8 @@ def verify_hitl_approval(
     # HITL-003: enforce approval expiry before verifying signature
     duration = approved_scope.get("approval_duration_seconds", 0)
     if duration:
-        if not isinstance(duration, (int, float)):
+        # bool is an int subclass, but a JSON boolean is not a numeric duration.
+        if not isinstance(duration, (int, float)) or isinstance(duration, bool):
             raise ValueError(
                 f"HITL approval.approved_scope.approval_duration_seconds must be "
                 f"numeric, got {type(duration).__name__}"
@@ -570,7 +571,14 @@ def verify_hitl_approval(
             approved_at = datetime.fromisoformat(approved_at_str.replace("Z", "+00:00"))
         except ValueError as e:
             raise ValueError(f"HITL approval has invalid approved_at: {e}") from e
-        if datetime.now(timezone.utc) > approved_at + timedelta(seconds=duration):
+        try:
+            expiry = approved_at + timedelta(seconds=duration)
+        except OverflowError as e:
+            raise ValueError(
+                f"HITL approval.approved_scope.approval_duration_seconds is out of "
+                f"range: {duration}"
+            ) from e
+        if datetime.now(timezone.utc) > expiry:
             raise ValueError(
                 f"HITL approval expired: approved_at={approved_at_str}, "
                 f"duration={duration}s"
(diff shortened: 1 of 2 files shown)


carloshvp (changes requested):
I reproduced two remaining boundary failures on head `3dcb0274378b47ee24b23f8679595890145cbc70`.

1. **Blocking: malformed base64url can still be accepted.** `base64.urlsafe_b64decode()` is permissive by default and silently discards characters outside the alphabet. Starting with a valid approval, this is accepted without any exception:

   ```python
   approval["approval_signature"] = "!!!!" + approval["approval_signature"]
   verify_hitl_approval(approval, manifest_id, key.public_bytes)
   ```

   The ignored prefix leaves the original 64 signature bytes, so verification succeeds. A value such as `"****"` is instead classified as `InvalidSignature`, not the documented `ValueError`. The new test only covers an input whose length happens to make the decoder raise. Please use strict alphabet validation, for example `base64.b64decode(..., altchars=b"-_", validate=True)`, and add a regression where illegal characters prefix an otherwise valid signature.

2. **Blocking: an oversized JSON integer still leaks `OverflowError`.** This reaches `timedelta(seconds=duration)` after passing the new numeric check:

   ```python
   approval["approved_scope"]["approval_duration_seconds"] = 10**20
   verify_hitl_approval(approval, manifest_id, key.public_bytes)
   # OverflowError: Python int too large to convert to C int
   ```

   Please validate the supported duration range or translate `OverflowError` to the helper's documented `ValueError`. It would also be worth excluding `bool`, which Python treats as an `int` even though a JSON boolean is not a numeric duration.

The focused delegation tests pass (`107 passed`), and the full suite passes (`1422 passed, 6 skipped, 1 xfailed`). Those suites currently do not cover the two cases above.

kingztech2019:
Thanks for the thorough review — both were real gaps, appreciate you reproducing them.

Fixed in 2e31c91:

- Switched `base64.urlsafe_b64decode()` to `base64.b64decode(sig + padding, altchars=b"-_", validate=True)`. Confirmed the permissive decode was the issue — `"!!!!" + <valid sig>` decoded to identical bytes as the untampered signature with no exception, so tampered-but-decodable input was silently accepted. `validate=True` now rejects it. Added a regression test reproducing your exact case.
- Wrapped the `timedelta(seconds=duration)` call in try/except, translating `OverflowError` to `ValueError` so oversized durations (e.g. `10**20`) hit the documented contract instead of leaking the stdlib exception. Also excluded `bool` from the numeric-duration check, since `isinstance(True, int)` is `True` but a JSON boolean isn't a valid duration — good catch.

Added 3 new tests (illegal-char-prefixed signature, oversized duration, bool duration). Full suite (1425 tests), mypy, ruff, and bandit all pass.

carloshvp (changes requested):
The two original findings are fixed on `2e31c916ac62e2771f023ed9bad193e41c1053df`, but I found two remaining malformed-input paths that are still accepted.

1. **Blocking: falsy malformed durations bypass the new type check.** The validation is inside `if duration:`, so it never examines `False`, `""`, `[]`, `{}`, or `None`. A validly signed approval containing any of those values is accepted by `verify_hitl_approval()`. For example:

   ```python
   scope = {**APPROVAL_SCOPE, "approval_duration_seconds": False}
   approval["approved_scope"] = scope
   approval["approval_signature"] = HitlApprovalSigner(kp).sign_approval(
       manifest_id=MID,
       approved_at=approval["approved_at"],
       approved_scope=scope,
       approver_id=approval["approver_id"],
   )
   verify_hitl_approval(approval, MID, kp.public_bytes)  # accepted
   ```

   This contradicts the stated rule that a JSON boolean is not a numeric duration and the issue's requirement to classify malformed duration shapes. Please run the type check before the truthiness/expiry branch. The default or explicit numeric `0` can still retain the existing no-expiry behavior.

2. **Blocking: `validate=True` does not restrict decoding to the base64url alphabet.** With `altchars=b"-_"`, `base64.b64decode()` accepts both `-_` and the standard-base64 `+/` characters. Replacing every `-` with `+` and `_` with `/` in an otherwise valid signature produces identical bytes and is still accepted:

   ```python
   approval["approval_signature"] = approval["approval_signature"].replace("-", "+").replace("_", "/")
   verify_hitl_approval(approval, MID, kp.public_bytes)  # accepted
   ```

   Please enforce the URL-safe alphabet or canonical unpadded representation before decoding, and add this equivalent-encoding regression alongside the illegal-character-prefix test.

I reran the prior reproductions and the new tests, then ran the focused delegation tests (`110 passed`) and full suite (`1425 passed, 6 skipped, 1 xfailed`). The current suites do not cover these two accepted malformed representations.

kingztech2019:
Both reproduced and fixed in fa9cc05.

- Moved the `approval_duration_seconds` type check above the `if duration:` truthiness gate. Confirmed `False` (and `""`, `[]`, `{}`, `None`) previously bypassed the check entirely and was silently treated as absent/zero duration — the type check now runs unconditionally, so all of those are rejected. The legitimate no-expiry behavior for an actual `0`/absent duration is unchanged.
- Whitelisted the URL-safe alphabet (`[A-Za-z0-9_-]`) before decoding. Confirmed `altchars=b"-_"` translates `-`/`_` to `+`/`/` *before* `validate=True` runs, so a signature with `+`/`/` substituted in place of `-`/`_` decoded to identical bytes and passed. The explicit whitelist closes that.

Added regression tests for both (comprehensive falsy-value parametrization for the duration case, a fixed non-random string for the alphabet case so the test doesn't depend on the fixture signature happening to contain `-`/`_`). Full suite (1431 tests), mypy, ruff, and bandit all pass.

carloshvp (dismissed):
Re-reviewed exact head `fa9cc05024d692a92c2f8d4bd280f5dcd89535f2`. The two findings from the prior review are fixed.

- `approval_duration_seconds` is now type-checked before the truthiness branch, so `False`, `None`, empty strings/containers, and other non-numeric values cannot silently become the no-expiry case. JSON booleans are rejected separately from integers.
- `approval_signature` now enforces the unpadded URL-safe alphabet before strict decoding, so illegal-character prefixes and standard-base64 `+`/`/` aliases cannot decode to the same accepted signature bytes.

The focused HITL suite passes (`62 passed`), the full suite passes (`1431 passed, 6 skipped, 1 xfailed`), and the substantive CI checks are green. The remaining failed policy check is the expected maintainer-review gate.

imran-siddique (approved):
Reviewed 372bfa87001c067a4cfe06f391d7ec375f7364a8. The validation rejects naive approval timestamps with ValueError before expiry comparison. The focused HITL suite passes all 70 tests after merging main, including date-only and naive timestamp regressions, valid signed timezone-offset controls, and main's approval-method signature binding. The timestamp regressions fail with TypeError before the repair. The merge resolution keeps the validated fields and includes approval_method in the signature preimage. CI is green apart from the maintainer-review gate.

I pushed the timestamp repair, so the repository's last-push rule still requires an independent reviewer before merge.

imran-siddique (commented):
Review/test follow-up on `372bfa87001c`.

The malformed-input contract now includes timezone-naive approvals, and the main-branch signed approval-method binding is retained. The full local suite and fresh hosted CI passed. No additional defect found in this diff. Since I pushed the repair, protected auto-merge still needs an eligible independent approval of the current head.

Full local `python -m pytest -q` (Windows, Python 3.12, existing review environment): **1487 passed, 6 skipped, 1 xfailed, 2 warnings in 24.47s**.

Fresh workflow evidence: [34012201062](https://github.com/agentrust-io/agent-manifest/actions/runs/34012201062) (success).

kingztech2019:
Thanks for pushing the repair directly, and for the detailed rationale on binding `approval_method` into the pre-image — that's a real gap (a validly-signed `software-key` approval being relabeled `hardware-key` post-signature without invalidating it) and a good catch independent of the two issues this PR started from.

Pulled the branch and re-ran the full suite locally rather than just taking the numbers as given: 1487 passed, 6 skipped, 1 xfailed — matches what you reported. No concerns on my end.

Looks like this needs a second maintainer's approval to clear branch protection since you pushed the latest commit — happy to wait on that.

PR #379, opened 2026-09-03
fix(plugins): verify_plugin_manifest_reference reads source_bundle before structural verification

### LINKED ISSUE #361: verify_plugin_manifest_reference reads source_bundle before structural verification
## Current behavior

`verify_plugin_manifest_reference()` is an exported result-returning verifier. It resolves a manifest referenced by an Agent Plugins bundle, checks the fetched-byte digest, parses the manifest, compares the manifest's signed `source_bundle` back to the local bundle, and then calls `verify_manifest()`.

At current `main` (`eb747f5fd610a8d7fa360e52faaea5db578e6b34`), the `source_bundle` comparison happens **before** the manifest verifier establishes schema/structure:

```python
source = manifest.get("source_bundle") or {}
expected_bundle_digest = bundle_reference_digest(path)
if (
    source.get("format") != "agent-plugins-1.0.0"
    or source.get("digest") != expected_bundle_digest
):
    return PluginReferenceResult(...)

verified = verify_manifest(verification_subject, context, revocation_store)
```

A fetched JSON/COSE manifest whose bytes exactly match the plugin's declared `manifest_digest` can still carry a truthy non-object `source_bundle`, for example a string or list. In that case `source.get(...)` raises `AttributeError` instead of producing a `PluginReferenceResult`.

Digest correspondence is evidence that the fetched bytes are the bytes referenced by the plugin. It is not structural authority for the contents of those bytes. The current ordering treats those as the same prerequisite.

## Expected invariant

Remote manifest content should not be interpreted through nested object operations until the structure those operations require has been established. A malformed referenced manifest should produce a fail-closed `PluginReferenceResult`, not a host-language exception.

## Classification boundary

I am intentionally **not** choosing whether a malformed `source_bundle` should be reported as `INVALID`, `MISMATCH`, or through another existing result. That classification is a project/API decision.

The narrower defect is independent of that choice: a result-returning verifier should not leak `AttributeError` from malformed referenced evidence.

## Suggested narrow directions

Either:

1. establish that `source_bundle` is an object before the local-bundle binding comparison; or
2. move the relevant structural prerequisite ahead of the binding read without changing the intended trust-ordering of full signature verification.

Whichever direction is chosen should preserve the existing distinction between:

- fetch/digest mismatch;
- malformed manifest structure;
- signed source-bundle mismatch;
- unavailable trust key/capability;
- successful authenticated binding.

## Regression matrix

At minimum:

- referenced JSON manifest with `source_bundle` = string/list/bool -> result, never exception;
- the same malformed shape in a COSE payload -> result, never exception;
- missing `source_bundle` -> existing mismatch behavior preserved;
- well-formed wrong bundle digest -> existing mismatch behavior preserved;
- valid referenced manifest/local bundle control -> existing VERIFIED path preserved.

No plugin schema, manifest wire format, signing pre-image, or trust-root policy change is proposed. This is structural prerequisite ordering at an exported verification boundary.

kingztech2019:
## What

`verify_plugin_manifest_reference()` now checks that `source_bundle` is an object before reading it, returning `PluginReferenceStatus.INVALID` instead of raising `AttributeError`.

## Why

Closes #361. `source = manifest.get("source_bundle") or {}` only guards against falsy values (`None`, `""`, `[]`, `{}`); a truthy non-object (a string, list, or bool) reaches `source.get("format")` unguarded. Digest correspondence between the fetched bytes and `manifest_digest` is evidence the bytes are the ones the plugin referenced — it is not structural authority for what those bytes contain, and the code was treating those as the same prerequisite.

This runs before `verify_manifest()`, on both the JSON and COSE fetch paths, so a manifest can reach it without ever needing a valid signature. `manifest` itself is already guaranteed to be a dict by both parsers (`_json_manifest` and `read_payload_manifest`/`_parse_payload` both raise `ValueError` subclasses otherwise, already caught) — `source_bundle` was the one unguarded nested field.

## Spec impact

None. SDK-only. No plugin schema, manifest wire format, signing pre-image, or trust-root policy change — purely structural prerequisite ordering at an exported verification boundary, per the issue's own scoping.

## Test plan

- [x] `pytest -v` passes (1407 passed, 6 skipped, 1 xfailed — no regressions; this branch predates #360's merge, hence the different total from that PR)
- [x] `bandit -r src/agent_manifest` passes
- [x] New tests: non-object `source_bundle` (string/list/bool) → `INVALID` with a message naming the field; missing `source_bundle` → existing `MISMATCH` preserved; well-formed-but-wrong-digest `source_bundle` → existing `MISMATCH` preserved. Exercised via the JSON path — the COSE path converges into the same line once its own parser has produced a dict, so the fix is shared code rather than something needing a duplicate test, noted in the test file rather than asserted without checking. 39/39 in `test_plugins.py`.
- [x] N/A — SDK-only, no spec change

## DCO

### DIFF
--- a/python/src/agent_manifest/_plugins.py
+++ b/python/src/agent_manifest/_plugins.py
@@ -354,6 +354,11 @@ def verify_plugin_manifest_reference(
         return PluginReferenceResult(PluginReferenceStatus.INVALID, str(exc))
 
     source = manifest.get("source_bundle") or {}
+    if not isinstance(source, dict):
+        return PluginReferenceResult(
+            PluginReferenceStatus.INVALID,
+            f"manifest source_bundle must be an object, got {type(source).__name__}",
+        )
     expected_bundle_digest = bundle_reference_digest(path)
     if (
         source.get("format") != "agent-plugins-1.0.0"
--- a/python/tests/test_plugins.py
+++ b/python/tests/test_plugins.py
@@ -364,6 +364,115 @@ def test_resolved_reference_verifies_manifest_and_bundle(tmp_path):
     assert result.manifest_result.signature_verified is True
 
 
+# --- malformed source_bundle must return a result, never raise - issue #361 ---
+#
+# The source_bundle check runs before verify_manifest(), on both the JSON and
+# COSE fetch paths (see verify_plugin_manifest_reference), so a manifest whose
+# bytes exactly match the declared manifest_digest can still carry a truthy
+# non-object source_bundle without ever needing a valid signature. Exercised
+# here via the JSON path; the COSE path converges into the exact same
+# `source = manifest.get("source_bundle") or {}` line once read_payload_manifest
+# has produced its dict, which _parse_payload already guarantees (it raises
+# CoseStructureError, a ValueError subclass, for anything else) - the fix is
+# shared code, not duplicated per format.
+
+
+def _manifest_with_source_bundle(root, source_bundle):
+    now = datetime.now(timezone.utc)
+    manifest = {
+        "manifest_id": "018f4a3b-2c1d-7e5f-a8b9-0d1e2f3a4b5c",
+        "agent_id": "spiffe://trust.example/plugin/demo",
+        "version": "0.1",
+        "issued_at": now.isoformat().replace("+00:00", "Z"),
+        "expires_at": (now + timedelta(days=30)).isoformat().replace("+00:00", "Z"),
+        "issuer": "spiffe://trust.example/issuer/ci",
+        "crypto_profile": "standard",
+        "source_bundle": source_bundle,
+        "artifacts": {
+            "system_prompt": {"hash": "sha256:" + "a" * 64},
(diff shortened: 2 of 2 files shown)


imran-siddique (approved):
Reviewed the source_bundle guard and its JSON/COSE convergence point on 1831727. Locally, the plugin-reference suite passes 39 tests. Restoring the pre-fix _plugins.py makes the three truthy non-object source_bundle cases fail with the original exception; the remaining 36 pass. The existing missing-object and well-formed mismatch results are preserved. Signed-off commit, substantive CI green, no unresolved review threads.

imran-siddique (approved):
Reviewed the source_bundle guard and its JSON/COSE convergence point on 1831727. Locally, the plugin-reference suite passes 39 tests. Restoring the pre-fix _plugins.py makes the three truthy non-object source_bundle cases fail with the original exception; the remaining 36 pass. The existing missing-object and well-formed mismatch results are preserved. Signed-off commit, substantive CI green, no unresolved review threads.

PR #384, opened 2026-09-04
fix(cert-chain): enforce pathLenConstraint (RFC 5280 §4.2.1.9)

rajnisht7:
## What

Enforce `pathLenConstraint` (RFC 5280 §4.2.1.9) in `verify_cert_chain`, which previously accepted chains that violate it.

## Why

`_cert_chain.py` is the shared X.509 chain verifier used by the SEV-SNP, TDX, and TPM attestation verifiers. It checks certificate validity, `BasicConstraints.ca`, and key usage on every issuer, but never read `BasicConstraints.path_length`. So a CA cert declaring "no CA may follow me" (`path_length=0`) could still have another CA below it in the chain, and the chain would verify anyway every other check on it passes individually. 

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Fixed
+
+- **[SDK]** `verify_cert_chain()` now enforces `BasicConstraints.path_length`
+  (RFC 5280 §4.2.1.9). An issuer that declares `path_length=N` but has more
+  than `N` CA certificates between it and the leaf previously verified
+  successfully every other individual check on it (validity period, `ca`
+  flag, key usage) passed on its own, so nothing else caught it. Self-issued
+  certificates (CA key rollover) are not exempted from the count; this is the
+  fail-closed direction relative to the full RFC algorithm.
+
 ### Deprecated
 
 - **[SPEC] Issuing v0.1 manifests ends 2026-11-30** (issue #315, phase 5). From that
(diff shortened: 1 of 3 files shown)


carloshvp (approved):
Reviewed exact head `c60fb0e56c78335954f3a7229e1efc8fe7ceff67` in a fresh detached checkout. The path-length check applies at every issuer, including intermediate CAs, with the correct zero/direct-leaf and exact-boundary behavior for the documented attestation-chain profile.

Validation: full Python suite `1417 passed, 6 skipped, 1 xfailed`; an independent matrix of 90 signed chains varying chain depth, constrained issuer position, and allowed depth also passed. The deliberately stricter treatment of self-issued rollover certificates is explicitly documented; this is not a claim of full RFC 5280 path-validation conformance. No blocking findings in this change. Substantive CI is green; the remaining failure is the maintainer-review policy gate.

imran-siddique (approved):
Reviewed path-length enforcement throughout the issuer chain, including its documented conservative treatment of self-issued intermediates. All 21 certificate-chain tests pass locally; restoring the previous implementation makes both new over-depth rejection cases fail.

imran-siddique (approved):
Reviewed path-length enforcement throughout the issuer chain, including its documented conservative treatment of self-issued intermediates. All 21 certificate-chain tests pass locally; restoring the previous implementation makes both new over-depth rejection cases fail.

PR #392, opened 2026-09-06
fix(trace): don't mask falsy non-dict artifacts/policy_bundle as missing

rajnisht7:
## What

Fixes ` _check_manifest_binding()` so a falsy non-dict `artifacts` or `policy_bundle` (`""`, `[]`, `False`, `0`) reports the correct structural warning instead of being silently treated as "no hash available."

## Why

`manifest.get("artifacts") or {}` folds any falsy value into `{}` before the `isinstance` check runs, so those cases were reported as `manifest_has_no_policy_bundle_hash` (missing) instead of `manifest_artifacts_not_an_object` / `manifest_policy_bundle_not_an_object` (malformed). Callers couldn't tell "hash absent" from "shape unreadable" for these inputs.

## Spec impact

none

## Test plan

- [x] `pytest -v` passes

## DCO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Unreleased
 
+### Fixed
+
+- **[SDK]** `_check_manifest_binding()` no longer masks a malformed
+  `artifacts` or `artifacts.policy_bundle` as merely absent. Truthy
+  non-objects (`"str"`, `[1]`, `True`) already correctly reported
+  `manifest_artifacts_not_an_object` / `manifest_policy_bundle_not_an_object`,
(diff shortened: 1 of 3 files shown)


imran-siddique (approved):
Reviewed 5fe43a61da82c5a7abeff8260532b7133780ba77. Missing or null artifacts retain the existing missing-hash warning; falsy values of the wrong type now produce the structural warning, including through evidence-pack verification. All 134 TRACE verification tests pass locally. Restoring the old implementation produces nine failures and 125 passes, confirming the new tests distinguish the defect. Hosted checks are green apart from the maintainer gate, with no held workflows or unresolved review threads.

PR #393, opened 2026-09-06
fix(delegation): non-empty Cedar constraints must be UNVERIFIABLE, not VALID

rajnisht7:
## What

`verify_delegation_chain()` no longer returns `VALID` when a hop has non-empty `scope_grant.constraints` it now raises `DelegationUnverifiable`, which `verify_manifest()` maps to UNVERIFIABLE.

## Why

Constraints are Cedar statements (spec 3.4.1), and this package has no Cedar parser/evaluator. `_check_scope_narrowing()` only did a string-set diff between parent and child constraints, so two hops carrying the exact same unparseable garbage string (nothing "dropped") verified as `VALID` end to end. Spec 3.4.1 / 5.2 explicitly require `UNVERIFIABLE` in this case, never `VALID`.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@
 
 ### Fixed
 
+- **[SDK]** `verify_delegation_chain()` no longer verifies a delegation chain
+  as `VALID` when a hop's `scope_grant.constraints` is non-empty. Constraints
+  are Cedar statements (spec 3.4.1); this package has no Cedar parser or
+  evaluator, so per spec such a chain MUST be reported `UNVERIFIABLE`, never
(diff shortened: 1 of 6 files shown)


imran-siddique (approved):
Reviewed 2ae9d59883bebd96970ccad39f25817bcb7a372d.

The verifier now preserves INVALID precedence for malformed or incorrectly signed later hops and reports unresolved Cedar constraints as UNVERIFIABLE only after checking the entire chain. The empty-constraint control remains VALID. Full local Python suite: 1,492 passed, 6 skipped, 1 expected failure. The new public exception is exported and documented; verify_manifest maps it explicitly.

PR #402, opened 2026-09-07
docs(spec): record the enforcement mode vocabulary crosswalk in 6.2.1

### LINKED ISSUE #344: docs: reconcile enforcement-mode vocabulary across Agent Manifest, cMCP, and TRACE
## Problem

`spec/agent-manifest-spec-v0.2.md:1582` makes the enforcement-mode cross-check normative:

> | Enforcement mode | `enforcement_mode` in TEE report | `policy_bundle.enforcement_mode` | MUST match. Conflict = attestation failure. |

and line 401 says the field "aligns with cMCP's `enforcement_mode` attestation field".

Neither is currently deliverable, for two independent reasons.

## 1. `VerificationContext` has no `enforcement_mode`

Installed `agent-manifest==0.11.1` and enumerated `VerificationContext.model_fields`:

```
system_prompt_hash, policy_bundle_hash, tool_catalog_hash, model_version,
rag_corpus_merkle_root, memory_snapshot_hash, audit_chain_root,
audit_chain_continuity, container_image_digest, challenge_nonce, purpose,
verifier_id, enforce_hitl, enforce_attestation, min_slsa_level, trusted_keys,
trusted_key_issuers, delegation_public_keys, approver_public_keys,
verified_transparency_entry_ids, verified_transparency_receipt_hashes,
transparency_evidence_manifest_id, require_transparency,
strict_artifact_verification, require_delegation, conformance_level
```

Twenty-six fields, no `enforcement_mode`. A trusted component has no way to supply the runtime value, so `verify_manifest()` cannot perform the comparison the spec requires. Reported downstream in agentrust-io/cmcp#576 by @athena-kanellatou, who found the missing cross-check on the cMCP side.

## 2. The two vocabularies do not reconcile

| agent-manifest | cMCP |
|---|---|
| `enforce` | `enforcing` |
| `advisory` | `advisory` |
| `audit-only` | *(none)* |
| *(none)* | `silent` |

`cmcp_runtime/audit/trace_claim.py:379` maps `{"enforcing": "enforce", "advisory": "advisory", "silent": "silent"}` for the TRACE claim, where `silent` is a legal value. There is no mapping onto this spec's set, and no value here means what cMCP means by `silent`.

So a cMCP deployment running `silent` has nothing correct to declare in `policy_bundle.enforcement_mode`. Picking `audit-only` would assert something different: TRACE §4.3 distinguishes these precisely, and in cMCP `silent` still evaluates policy and records every would-have-denied decision while suppressing operational log lines. That is not the same as auditing without enforcing.

## What needs deciding

1. **Add `enforcement_mode` to `VerificationContext`** and perform the §6.2 comparison in `verify_manifest()`, treating a conflict as an attestation failure as the spec says.
2. **Reconcile the vocabularies.** Either add a value meaning what `silent` means, or state a normative mapping, or narrow the spec's claim that the field aligns with cMCP's. Doing (1) without (2) produces a verifier that rejects a legitimate `silent` deployment or accepts a mislabelled one.

The second is the real decision and it should come first, because it determines what the field can hold.

## Related

- agentrust-io/cmcp#576, the downstream report
- cmcp#342, whose "extend to cover enforcement_mode attestation binding" follow-up was closed with the hash-binding work

zohebk8s:
## What

Adds an informative section 6.2.1 recording how the three enforcement mode vocabularies correspond, and points the alignment sentence in 3.2.2 at it.

## Why

Closes #344. Section 6.2 requires the cMCP attested enforcement mode and `artifacts.policy_bundle.enforcement_mode` to match, and 3.2.2 says the two fields align, but neither says which name denotes which state. Three vocabularies are in play:

| State | Agent Manifest | cMCP runtime | TRACE claim |
|---|---|---|---|
| Decisions applied | `enforce` | `enforcing` | `enforce` |
| Decisions surfaced, call proceeds | `advisory` | `advisory` | `advisory` |
| Decisions recorded, reporting suppressed | `audit-only` | `silent` | `silent` |

Until now the only written statement of the correspondence was a mapping table inside a consumer, agentrust-io/cmcp#584, so a second implementer had to rediscover it or guess. On cmcp `main` at `5b63642`, 346 of its 385 files decode as text and the string `audit-only` appears in exactly one of them.

The section records the correspondence without asserting the states are identical. `silent` suppresses operational reporting while still recording each decision, so `silent` satisfies `audit-only` and `audit-only` does not imply `silent`. Keeping that difference explicit is what the review comment asked for; the direction of the implication is my reading and is flagged below.

## Spec impact

Section 6.2.1 is new and informative. Section 3.2.2 gains a pointer clause. No RFC 2119 keyword is added, no requirement changes, and no conformance test is affected, so no sponsor is required under GOVERNANCE. `CHANGELOG.md` updated under Unreleased.

One sentence needs an author's eye. The specification has never defined what its three modes mean; `enforce`, `advisory` and `audit-only` appear only in two schema blocks and the 3.2.2 sentence. To keep the suppressed-reporting difference explicit I had to write what `audit-only` means, which makes this the first written definition of it in the spec:

> `audit-only` describes evaluating and recording without applying, and says nothing about whether the decision is reported.

That reading follows the argument in the issue, but it is an interpretation rather than a transcription. Happy to reword it.

## Test plan

Run on this branch, not assumed from the fact that only documentation changed.

- [x] `pytest -v` passes: 1481 passed, 6 skipped, 1 xfailed
- [x] `mypy src/agent_manifest` passes: no issues found in 26 source files
- [x] `ruff check src/ tests/ --select E,F,W --ignore E501` passes (the CI invocation, ruff 0.16.6 as pinned in `requirements/dev.txt`): all checks passed
- [x] New or updated tests cover the change: not applicable, no normative text changed and no code touched

## DCO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Unreleased
 
+### Added
+
+- **[SPEC] Section 6.2.1 records the enforcement mode vocabulary crosswalk** (issue #344).
+  The manifest names the three policy enforcement states `enforce`, `advisory` and
+  `audit-only`, the cMCP runtime names them `enforcing`, `advisory` and `silent`, and the
+  TRACE claim schema uses a third set, `enforce`, `advisory` and `silent`. Section 6.2
+  requires the manifest and the attested runtime mode to match but never said which name
+  denoted which state, so the only written statement of the correspondence sat inside a
+  consumer (agentrust-io/cmcp#584). The new subsection is informative, carries no RFC 2119
+  keyword, and adds no requirement beyond the existing row. It records the correspondence
+  without asserting that `silent` and `audit-only` are the same state. Reported by
+  Imran Siddique.
+
 ### Fixed
 
 - **[SECURITY][CLI]** `manifest verify` gained `--crl-trusted-key PATH` to
--- a/spec/agent-manifest-spec-v0.2.md
+++ b/spec/agent-manifest-spec-v0.2.md
@@ -398,7 +398,7 @@ Deliberately open. Whether Level 2 conformance should require `result: passed`, 
 }
 ```
 
-The policy bundle hash covers the complete Cedar policy set, including all policy templates and entity schemas. The `enforcement_mode` field is normative - a verifying party MUST reject a manifest whose `enforcement_mode` is `advisory` when the context requires `enforce`. This field aligns with cMCP's `enforcement_mode` attestation field.
+The policy bundle hash covers the complete Cedar policy set, including all policy templates and entity schemas. The `enforcement_mode` field is normative - a verifying party MUST reject a manifest whose `enforcement_mode` is `advisory` when the context requires `enforce`. This field aligns with cMCP's `enforcement_mode` attestation field; section 6.2.1 records how the cMCP runtime and TRACE claim mode names correspond to these values. <!-- CHANGED: ISSUE-344 - point the alignment claim at the crosswalk that substantiates it -->
 
 For `policy_language: composite`, the `hash` field MUST be a Merkle root over the hashes of each sub-bundle, sorted by policy language identifier in lexicographic order (`cedar`, `rego`, `yaml-agt`). Each sub-bundle MUST be hashed independently using the same hash algorithm as the manifest. The `agt_version` field MUST reference the AGT version used to assemble the composite bundle, even if individual sub-bundles were produced by other tools.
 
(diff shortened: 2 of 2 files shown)


imran-siddique (approved):
Reviewed df0a026ff91a187da75544174e5a9d38e10e0ab9.

This implements the informative mapping requested in #344. I checked the table against cMCP's centralized manifest mapping and TRACE claim emitter. The text preserves the distinction between recording without enforcement and suppressing operational reporting, explains the caller-side translation, and leaves a normative vocabulary extension open. Accepting the stated audit-only interpretation for this informative crosswalk. No executable behavior changes; current hosted tests, lint and type checks pass.

PR #403, opened 2026-09-07
fix: require issuer for v0.2 manifests (was silently optional like legacy v0.1)

rajnisht7:
## What

Scopes the "missing issuer is OK" schema exception to v0.1 manifests only, so v0.2 manifests without issuer correctly fail verification instead of coming back VALID.

## Why

`_strict_schema_violations()` in `_verify.py` was filtering out the missing-issuer schema error for every manifest version, with no version check. The exception was only ever meant for legacy v0.1 records (which predate the issuer field). Since the v0.2 spec makes issuer REQUIRED, and it's used elsewhere for signing-key authorization, a v0.2 manifest with issuer stripped out could carry a valid COSE signature and still verify as VALID silently dropping that authorization check.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@
 
 ### Fixed
 
+- **[SECURITY][SDK]** `_strict_schema_violations()` tolerated a missing
+  top-level `issuer` claim for **any** manifest version, not just v0.1. The
+  exception was added for legacy v0.1 records, which predate the `issuer`
+  field (CHANGELOG: "legacy v0.1 issuer omission remains compatible"), but
+  the filter never checked which version was being verified. The v0.2 spec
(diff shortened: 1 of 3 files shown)


imran-siddique (approved):
Reviewed 367d0c0dd3f4d2971c2d5bb36ca1a369d9fc5ff3.

The issuer omission exception is now limited to legacy version 0.1. Full local suite: 1,495 passed, 6 skipped, 1 expected failure. I reverted only this condition in an isolated worktree: both new v0.2 tests then failed because verification returned VALID, while the v0.1 compatibility control passed. Restored the exact reviewed tree afterward. This establishes the schema rejection without breaking the legacy path; current hosted CI is passing apart from the review gate.

PR #407, opened 2026-09-08
Fix: Bound POST /verify (JSON) body size and field cardinality

### LINKED ISSUE #383: `POST /verify` (JSON) has no body-size or field-cardinality cap, unlike the COSE `/verify` endpoint
## Spec section

N/A

## Problem

`_verify.py` implements two verification endpoints:

- `POST /verify` (COSE, `verify_cose`) : accepts a raw CBOR body. This handler explicitly bounds the request before parsing it: it checks `Content-Length` against `MAX_COSE_ENVELOPE_BYTES` (1 MiB), and then also caps the actual byte stream as it's read, with the comment "Content-Length is a claim, not a guarantee." A body over the limit gets `413 ENVELOPE_TOO_LARGE`. The docstring calls this out as a deliberate security property, separate from auth/rate-limiting, which the same docstring explicitly says are left to deployment.

- `POST /verify` (JSON, `verify_post`) " takes a `VerifyRequest` Pydantic model directly as the parameter. FastAPI parses the JSON body into this model with no byte-size cap before parsing, and none of `VerifyRequest`'s dict/set fields (`trusted_keys`, `trusted_key_issuers`, `delegation_public_keys`, `approver_public_keys`, `verified_transparency_entry_ids`, `verified_transparency_receipt_hashes`) have any cardinality limit.

A caller can send an arbitrarily large JSON body e.g. a `trusted_keys` dict with millions of entries and it will be fully buffered and validated by Pydantic before `verify_manifest` ever runs. That's unbounded memory/CPU work per request, on an endpoint that (per its own docstring) is meant to sit in front of untrusted relying parties.

Also there's no global body-size middleware anywhere in the codebase that would catch this either.

## Proposed change

Mirror the COSE endpoint's existing pattern on the JSON endpoint, so both `/verify` routes have the same defense-in-depth posture:

1. **Bound the body before parsing.** Change `verify_post` to take the raw `Request` (like `verify_cose` already does), stream-cap it against a size limit, then hand the bytes to `VerifyRequest.model_validate_json(...)`. Reuse `MAX_COSE_ENVELOPE_BYTES` (rename to something endpoint-neutral, e.g. `MAX_VERIFY_BODY_BYTES`, since it's now shared) instead of inventing a second constant.

2. **Cap dict/set cardinality on `VerifyRequest`.** Add a single reusable Pydantic `field_validator` that rejects any of the six unbounded fields above once entry count passes a fixed ceiling (e.g. 10,000 far above what any real deployment needs, since these maps exist to hold keys for the hops/approvers/keys actually referenced by one manifest). This catches a small-request-with-huge-collection case that a byte cap alone might miss if someone sends short key strings.

3. **New error code** for the oversized case, `413` with `ErrorResponse(error_code="VERIFY_REQUEST_TOO_LARGE", ...)`  same shape as `ENVELOPE_TOO_LARGE`, just endpoint-neutral naming since both routes will use it.

4. `GET /verify` is unaffected, it only takes a `manifest_id` query param, no body.

Sketch:

```python
MAX_VERIFY_BODY_BYTES = 1 << 20  # was MAX_COSE_ENVELOPE_BYTES, now shared
MAX_VERIFY_COLLECTION_ENTRIES = 10_000

class VerifyRequest(BaseModel):
    ...
    @field_validator(
        "trusted_keys", "trusted_key_issuers", "delegation_public_keys",
        "approver_public_keys", "verified_transparency_entry_ids",
        "verified_transparency_receipt_hashes",
    )
    @classmethod
    def _cap_collection_size(cls, v):
        if len(v) > MAX_VERIFY_COLLECTION_ENTRIES:
            raise ValueError(
                f"exceeds {MAX_VERIFY_COLLECTION_ENTRIES}-entry limit"
            )
        return v

@router.post("/verify", response_model=VerificationResult)
async def verify_post(request: Request) -> VerificationResult:
    body = bytearray()
    async for chunk in request.stream():
        body.extend(chunk)
        if len(body) > MAX_VERIFY_BODY_BYTES:
            raise HTTPException(
                status_code=413,
                detail=ErrorResponse(
                    error_code="VERIFY_REQUEST_TOO_LARGE",
                    error_message=f"Request body may not exceed {MAX_VERIFY_BODY_BYTES} bytes.",
                ).model_dump(),
            )
    parsed = VerifyRequest.model_validate_json(bytes(body))
    ...  
```

## Impact

- Backward compatible: yes, only oversized/abusive requests are newly rejected; every existing valid request stays under both new limits by a wide margin.
- Affects conformance level(s): none. This doesn't touch manifest structure, signing, or verification semantics only transport-level handling of the JSON `/verify` request, which section 7.2 already places outside conformance.
- Conformance tests that need updating: none.
- Regulatory mapping impact: none.

## Alternatives considered

- **Leave it to deployment (reverse proxy body-size limits).** This is the SDK's own stated policy for auth/rate-limiting on both `/verify` routes, but the COSE handler already chose *not* to rely on it for body size specifically, with an explicit comment on why ("Content-Length is a claim, not a guarantee", a proxy limit has the same trust problem). Leaving the JSON route unprotected while the COSE route isn't is the inconsistency this issue is about, so this alternative doesn't resolve it.
- **Starlette/FastAPI body-size middleware instead of manual streaming.** Would work, but would diverge from the pattern already established in `verify_cose`, and off-the-shelf middleware doesn't get the "cap the stream, not just the declared header" behavior for free either. Keeping both endpoints on the same hand-rolled pattern is easier to review and keep in sync.

Happy to submit a PR for this fix if the approach above looks right

rajnisht7:
## What

Adds the same body-size and collection-cardinality limits to `POST /verify` (JSON) that `POST /verify/cose` already has.

## Why

`POST /verify` took a `VerifyRequest` Pydantic model directly, so FastAPI would fully buffer and validate an arbitrarily large JSON body e.g. a `trusted_keys` dict with millions of entries before any verification logic ran. `POST /verify/cose` already guards against this; this endpoint didn't. 

Closes #383.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,43 @@
   and `attach_unprotected()` converts any residual `CBOREncodeError` to
   `CoseStructureError` as a second layer.
 
+- **[SECURITY][SDK]** `POST /verify` (JSON) had no body-size or field-cardinality
+  cap, unlike `POST /verify/cose` (issue #383). `verify_cose` bounds the CBOR
+  body against `Content-Length` and then against the actual byte stream
+  before parsing it ("`Content-Length` is a claim, not a guarantee"), but
+  `verify_post` took a `VerifyRequest` Pydantic model directly, so FastAPI
+  fully buffered and validated an arbitrarily large JSON body - e.g. a
+  `trusted_keys` dict with millions of entries - before `verify_manifest`
+  ever ran. `verify_post` now takes the raw request, stream-caps the body at
+  `MAX_VERIFY_BODY_BYTES` (the renamed, now endpoint-neutral
+  `MAX_COSE_ENVELOPE_BYTES`, 1 MiB, shared by both `/verify` routes) before
+  handing the bytes to `VerifyRequest.model_validate_json()`, and returns
+  `413 VERIFY_REQUEST_TOO_LARGE` over the limit, mirroring `ENVELOPE_TOO_LARGE`.
+  `VerifyRequest`'s seven dict/set fields (`trusted_keys`, `trusted_key_issuers`,
+  `delegation_public_keys`, `approver_public_keys`,
+  `verified_transparency_entry_ids`, `verified_transparency_receipt_hashes`,
+  `verified_attestation_manifest_hashes` - the last of which the original
+  report's field list omitted, but which has the identical unbounded
+  `set[str]` shape and reaches `VerificationContext` the same way)
+  now also reject more than `MAX_VERIFY_COLLECTION_ENTRIES` (10,000) entries,
+  which catches a small body hiding an oversized collection behind short
+  keys/values that a byte cap alone could miss. `GET /verify` is unaffected;
+  it takes no body. Taking a raw `Request` instead of a `VerifyRequest`
+  parameter also meant FastAPI's own automatic `Content-Type` gate no
+  longer ran for this route, so `verify_post` would parse a JSON body
+  submitted under any (or no) `Content-Type` - `text/plain`, a bogus
+  subtype, form-encoded - contradicting the `application/json`-only
+  contract the endpoint's own OpenAPI schema advertises. `verify_post` now
+  checks `Content-Type` explicitly, using the same rule FastAPI itself uses
+  for a Pydantic-model body parameter (`application/json`, a `+json`
+  vendor subtype, or no header at all; anything else is `415
+  UNSUPPORTED_MEDIA_TYPE`), before the body is read. Separately, both
+  `verify_post` and `verify_cose` checked the accumulated body against
+  `MAX_VERIFY_BODY_BYTES` *after* appending each stream chunk rather than
+  before, so the buffer could be grown past the cap by up to one chunk's
+  size before being rejected; both now check before appending, so the
+  buffer itself is never intentionally grown past the limit.
+
 - **[SECURITY][SDK]** `_strict_schema_violations()` tolerated a missing
   top-level `issuer` claim for **any** manifest version, not just v0.1. The
   exception was added for legacy v0.1 records, which predate the `issuer`
--- a/python/src/agent_manifest/_verify.py
+++ b/python/src/agent_manifest/_verify.py
@@ -13,6 +13,7 @@ The FastAPI router wires the engine to HTTP.
 """
 from __future__ import annotations
 
+import email.message
 import hashlib
 import hmac
 import uuid
@@ -20,7 +21,7 @@ from datetime import datetime, timezone
 from enum import Enum
 from typing import Any, Callable, Optional, Union
 
-from pydantic import BaseModel, Field, TypeAdapter
+from pydantic import BaseModel, Field, TypeAdapter, field_validator
 
 from ._audit_continuity import AuditCheckpoint, verify_continuity
 from ._cose import (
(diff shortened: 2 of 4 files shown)


rajnisht7:
The failing test (`test_followup_tutorial[operations/monitoring.md-expected0]`) is unrelated to this PR it's a pre-existing dependency-lock issue.

`prometheus-client` is declared in `pyproject.toml`'s `dev` extra but was missing from `requirements/dev.in` `requirements/dev.txt`. 

Opened a PR for the same #408

imran-siddique (dismissed):
Approving. Three of the things in this PR were not asked for and each of them is the reason to take it rather than a bonus on top.

**The seventh field.** The report named six. `verified_attestation_manifest_hashes` has the identical unbounded `set[str]` shape, sits directly below two fields that were named, and reaches `VerificationContext` by the same path. Capping six and leaving the seventh open would have moved the attack rather than closed it, and your comment says exactly that in the code where the next reader needs it. Fixing the instances the reporter missed is the right instinct on a bounds defect, because the bound is only worth what its weakest field is.

**The regression you caught in your own change.** Swapping a `VerifyRequest` parameter for a raw `Request` removes FastAPI's automatic `Content-Type` gate, so the route would have parsed a JSON body submitted as `text/plain`, as a bogus subtype, or with no header at all, contradicting the `application/json` contract its own OpenAPI schema advertises. That is invisible in a diff review: the removed behaviour has no line in the patch. Restoring it explicitly with FastAPI's own rule (`application/json`, a `+json` vendor subtype, or absent) rather than a stricter homemade one is the right call, because a stricter gate would have been a silent behaviour change for existing callers.

**The pre-existing bug you found looking for the first one.** Both routes checked the accumulated buffer against the cap *after* appending each chunk, so the buffer could be grown a full chunk past the limit before rejection. Checking before appending is correct and it fixes `verify_cose`, which was not in scope. A size cap enforced after the allocation is not a size cap.

The 1 MiB byte cap and the 10,000-entry collection cap answer different attacks and both are needed: the byte cap alone misses a small body carrying a million short keys, and the entry cap alone misses one enormous value. Renaming `MAX_COSE_ENVELOPE_BYTES` to the endpoint-neutral `MAX_VERIFY_BODY_BYTES` and sharing it across both routes keeps them from drifting apart again, which is how this gap opened in the first place.

On CI: every one of the ten failing lines across the five platform jobs is `test_docs_first_manifest.py::test_followup_tutorial[operations/monitoring.md]` failing on `ModuleNotFoundError: No module named 'prometheus_client'`. That is #408, your other PR, now merged, and it was failing on `main` and on every open PR in this repository. Nothing in your diff caused a failure. I am closing and reopening this so CI recomputes against a `main` that has the lockfile fix, since a re-run would replay the original merge commit and keep the stale base.

rajnisht7:
CI is failing on `fuzz_cose`, it is because `attach_receipt`/`attach_attestation`/`attach_approvals` can crash on re-encode if the unprotected header contains malformed CBOR value (cbor2's internal break-marker sentinel), as it wasn't validated before being copied back out. 

Opened a PR for the same #411

carloshvp (dismissed):
Approved on exact head 590827a0d7409b7ec2f04bbf192e116c97000585.

Independent validation in an isolated checkout:

- Full Python suite: 1,581 passed, 6 skipped.
- mypy: no issues in 26 source files.
- Bandit: clean.
- git diff --check: clean.
- Baseline regression: current main accepts and parses the over-1-MiB JSON body before returning 404; this head rejects the same request with 413 VERIFY_REQUEST_TOO_LARGE.
- Reviewed transport and boundary cases for actual stream length, lying Content-Length, exact-size bodies, JSON media types, malformed JSON, OpenAPI/error-shape compatibility, and all seven caller-supplied collections.

The failing ClusterFuzzLite job is a separate current-main CBOR sentinel crash. Its log reproduces the CBOREncodeError fixed by #411, not a defect introduced by this diff.

imran-siddique:
Your #411 merged as `e49b445e`, and that conflicts this branch. Everything below is so the rebase costs you one command rather than an investigation.

**Rebase onto `e49b445ebe923aa802dc336105a407dd8c642cd3`.**

**The only conflict is `CHANGELOG.md`**, one hunk, and both sides are kept. Your `POST /verify` entry and the `attach_receipt()`/`attach_attestation()`/`attach_approvals()` entry from #411 were both written at the top of `### Fixed`, so git cannot pick an order. I put #411's first and yours second; the reverse is equally fine. No source file conflicts.

**What I got resolving it locally**, so you know what to expect rather than discovering it:

- `pytest tests/`: **1585 passed, 6 skipped**.
- Resulting diff against main: 4 files, +626/-27, `CHANGELOG.md`, `_verify.py`, `test_cose_endpoint.py`, `test_verify_post_bounds.py`.
- `_reject_cbor_sentinels` is present in `_cose.py` after the merge, 7 occurrences.

**That last point is the one worth having.** `Fuzz changed code` has been red on this PR and it was never your code. The failing step is `Run fuzzers`; `fuzz_canonicalize` and `fuzz_attestation_parsers` both pass and `fuzz_cose` dies with `CBOREncodeError: cannot encode type <class 'object'>`. That is the break-marker sentinel defect, it was on main, and **#411 is the fix**. Once you rebase past `e49b445e` that check should go green without you touching it. The same red is sitting on #373 for the same reason.

One other thing that was masking this: the earlier `Test` failures across five platform jobs were `ModuleNotFoundError: No module named 'prometheus_client'`, which was #408's packaging defect rather than anything here. That is already fixed on main and cleared on the last run.

My approval from 8 September stands and survives the rebase. This is bookkeeping, not a second review round.

imran-siddique (approved):
Re-approving. This is the same review, not a second round.

You rebased onto `e49b445e` exactly as asked and squashed to one commit. GitHub dismissed my approval and @carloshvp's automatically when the branch was rewritten; neither of us withdrew anything, and nothing about the change needed re-arguing.

Checked that the rebase preserved what the review covered rather than assuming it: still +626/-27 across the same four files, and at `c7c9ddbc` the guards are all present, `MAX_VERIFY_BODY_BYTES` in nine places, `MAX_VERIFY_COLLECTION_ENTRIES` in four, plus `VERIFY_REQUEST_TOO_LARGE` and `UNSUPPORTED_MEDIA_TYPE`.

One thing worth closing out, since it sat on this PR for days and was never yours. `Fuzz changed code` is now green. It was failing on `fuzz_cose` with `CBOREncodeError: cannot encode type <class 'object'>`, which was the break-marker sentinel defect living on main. Your own #411 fixed it, and rebasing past `e49b445e` picked that up. Same story for the earlier `ModuleNotFoundError: No module named 'prometheus_client'` across five platform jobs, which was #408's packaging defect.

So both of the reds that made this PR look broken belonged to main. Thanks for rebasing without asking me to justify it.

PR #408, opened 2026-09-08
fix(deps): add missing prometheus-client to dev lockfile

rajnisht7:
## What

Adds `prometheus-client` to `requirements/dev.in` and regenerates `requirements/dev.txt`.

## Why

`pyproject.toml`'s `dev` extra declares `prometheus-client>=0.20,<1` (needed for the `operations/monitoring.md` tutorial test), but `requirements/dev.in` never had it so `requirements/dev.txt`, the file CI actually installs via `pip install --require-hashes`, was missing it too.

Due to this `test_followup_tutorial[operations/monitoring.md-expected0]` to fail with `ModuleNotFoundError: No module named 'prometheus_client.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO

### DIFF
--- a/requirements/dev.in
+++ b/requirements/dev.in
@@ -19,6 +19,7 @@ click>=8.1
 mypy>=1.10
 bandit[toml]>=1.7
 pip-audit>=2.7
+prometheus-client>=0.20,<1
 ruff
 pyyaml
 setuptools>=83.0.0
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1040,6 +1040,10 @@ pluggy==1.6.0 \
     # via
     #   pytest
     #   pytest-cov
+prometheus-client==0.26.0 \
+    --hash=sha256:04a91bcf94e2cf74a44a1a874d651a2e853ed354b6e822f3b7487751465d5c2b \
+    --hash=sha256:fa93d06737aa02bacd05794768508bb97d2fbee28cb3bca04eaae92f0ca953d6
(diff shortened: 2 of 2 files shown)


imran-siddique (approved):
Merging. This is the repository's red, not a nicety: `test_docs_first_manifest.py::test_followup_tutorial[operations/monitoring.md]` is failing on `ModuleNotFoundError: No module named 'prometheus_client'` on `main` and on every open PR here, across all five platform jobs. I checked main's own run rather than inferring it from a PR.

The diagnosis in your body is exactly right and worth restating because the failure mode is easy to reintroduce. `pyproject.toml`'s `dev` extra declares `prometheus-client>=0.20,<1`, `requirements/dev.in` never listed it, and CI installs `requirements/dev.txt` with `--require-hashes`. So the declared development dependency and the installed one are two different facts, and nothing in the pipeline compares them: a dependency can be declared and simply never installed, and the only symptom is a test that imports it.

The fix follows the repository's own pattern rather than shortcutting it: the constraint goes in `dev.in`, `dev.txt` is regenerated with both hashes and the `# via -r requirements/dev.in` provenance line kept. A hand-added pin without a regenerated hash block would have failed `--require-hashes` on the next resolve.

Worth naming for whoever picks this up next: #407 is yours too, and its five red `Test` jobs are entirely this, nothing to do with its own diff. That is the cost of a shared failure sitting under a queue, and it is why this one goes first.

PR #415, opened 2026-09-11
fix: HITL verifier no longer rejects VALID approvals just because an earlier approval in the array failed  

rajnisht7:
## What

`verify_manifest()` now marks `hitl_record` as `APPROVED` if **any** approval in `hitl_record.approvals` is valid, unexpired, and sufficient instead of rejecting the whole record the moment it hits the *first* bad approval.

## Why

Spec 5.3 says a `VALID` result only needs **at least one** good approval. The verifier's loop `break`ed on the first expired/invalid/unverifiable/insufficient approval and never looked at the rest of the array. So `[expired_approval, valid_approval]` was wrongly rejected as `EXPIRED`, even though the second approval was perfectly valid. Same for an invalid or unverifiable approval placed before a good one. (HITL-004)

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,22 @@
 
 ### Fixed
 
+- **[SDK]** `verify_manifest()` returned `MISMATCH`/`EXPIRED`/`INVALID`/
+  `UNVERIFIABLE` for `hitl_record` when *any* approval in `hitl_record.approvals`
+  failed, even if a later approval in the same array was present, valid,
+  unexpired, and sufficient for the declared risk tier. Spec 5.3 only
+  requires *at least one* approval to satisfy every condition; it does not
+  require every approval in the array to. The verifier's loop broke out on
+  the first expired, malformed, unverifiable, or insufficient approval it
+  encountered and never evaluated the approvals after it, so an operator
+  who appended a fresh re-approval ahead of (or alongside) an old, expired,
+  or otherwise stale one had the whole record incorrectly rejected. The
+  loop now evaluates every approval and only short-circuits once one is
+  found that clears every check; if none do, the previous priority order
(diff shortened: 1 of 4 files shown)


imran-siddique (approved):
Approving. Measured against the spec text rather than the PR body, because this widens what the verifier accepts and that direction earns the check.

**The rule this is held to**, `spec/agent-manifest-spec-v0.2.md` section 5.3, in the list of conditions a `VALID` result requires:

> If `enforce_hitl` is `true`, at least one HITL approval is present, valid, not expired, and meets the `approval_method` requirement for the declared `risk_tier`

"At least one" is unambiguous, so breaking out of the loop on the first bad entry was a defect, and `[expired_approval, valid_approval]` returning `EXPIRED` was wrong under the spec we publish.

**What I checked beyond the headline**, since a loosened check is only safe if the failure path survives it:

- Each mode is still recorded separately, `any_expired`, `approval_insufficient`, `approval_unverifiable`, `approval_invalid`, so a record where *every* approval is bad still reports which way it failed rather than collapsing to one reason. That distinction is the reason this is a fix and not a weakening.
- The defensive fallback for a non-empty `approvals` where every entry fell through is still there.
- Unparseable `approved_at` still fails safe as expired (HITL-001). That is the line a later tidy-up would most easily drop, and it is doing real work.
- Level 2 still refuses `software-key`, and still refuses anything other than `hardware-key` for `high` and `critical` risk tiers, per entry rather than per record.

**Merge state.** #407 landed in `_verify.py` a few minutes before this, and the trial merge onto current main is clean with no conflicts in either that file or `CHANGELOG.md`. The merged tree runs 1572 passed, 10 skipped.

PR #417, opened 2026-09-12
Fix empty trusted_key_issuers semantics

### LINKED ISSUE #413: Empty trusted_key_issuers mapping bypasses issuer authorization
## What

`_signature_key_issuer_mismatch()` in (`_verify.py:412`) treats an empty `trusted_key_issuers` mapping as "no check configured" and returns `None` unconditionally:

```python
def _signature_key_issuer_mismatch(
    manifest: dict[str, Any],
    key_id: str,
    trusted_key_issuers: dict[str, list[str]],
) -> Optional[MismatchDetail]:
    """Return a mismatch when a trusted key is not authorized for the issuer."""
    if not trusted_key_issuers:
        return None
    ...
```

`POST /verify`'s own docstring describes `trusted_key_issuers` as fail-closed:

> `trusted_key_issuers` maps each trusted key_id to the issuer SPIFFE URIs authorized to sign manifests with that key. **When supplied, key-to-issuer authorization is fail-closed.**
> (`_verify.py:221-222`, `VerifyRequest`)

`VerifyRequest.trusted_key_issuers` flows unchanged into `VerificationContext.trusted_key_issuers` (`_verify.py:1725`), which is what `_signature_key_issuer_mismatch()` reads at both call sites, the COSE path (`_verify.py:855`) and the v0.1 path (`_verify.py:925`). Both are affected identically.

## Reproduction

Conformance vector `AM-VEC-024` (generated via the existing `tests/vectors/generate.py` harness, same shape as `AM-VEC-022`/`AM-VEC-023`). The manifest, signature and signing key are byte-identical to `AM-VEC-001`. Only `context.trusted_key_issuers` differs. `AM-VEC-024` sets it to `{}`.

```
$ pytest tests/test_vectors.py -k "AM-VEC-024" -v
AssertionError: AM-VEC-024: expected MISMATCH, got VALID
```

Full suite for context: `1 failed, 91 passed`. The remaining vectors pass.

## Why this appears inconsistent with the current contract

- `trusted_keys` and `approver_public_keys` on the same `VerifyRequest` are both documented and implemented as fail-closed. An absent/empty yields `UNVERIFIABLE`. `VALID` requires a resolved trusted key.
- `trusted_key_issuers` carries the same fail-closed documentation. The empty mapping takes the early return at line 418 before the authorization logic below it runs.
- The authorization logic below the early return correctly rejects a known key without authorization for the declared issuer, as demonstrated by `AM-VEC-022` and `AM-VEC-023`.
- Spec 5.3.3 (`docs(spec): require issuer key authorization`, #367) requires a verifier to resolve and authorize the signing key for the declared issuer before returning `VALID`, through a trust path independent of the manifest itself. An empty `trusted_key_issuers` mapping leaves that authorization unresolved.
- `git blame` puts `_signature_key_issuer_mismatch()` at 2026-07-07 (Carlos Hernandez), predating #367 (merged ~2026-08-30/09-01). The empty-mapping short-circuit therefore predates the current 5.3.3 requirement. Flagging in case I'm missing the intended semantics.

## Question

Is `if not trusted_key_issuers: return None` intentional (e.g., a deliberate opt-out for callers with no issuer-authorization data)?
Or should an empty mapping be treated as "authorization not established" the same way empty `trusted_keys` is? If the latter, `AM-VEC-024` is ready as the regression vector once the call site's fail-closed behavior is decided.

I can open a PR for the fix once the intended semantics are confirmed.

litzki-systems:
Fixes #413

`trusted_key_issuers=None` now represents an unset issuer authorization policy, while an explicitly empty mapping authorizes no issuers and therefore fails closed.

Adds regression vector AM-VEC-024 and updates the changelog accordingly.

Existing behavior for omitted `trusted_key_issuers` remains unchanged.

## Implementation and verification

This PR distinguishes between an unset issuer authorization policy and an explicitly empty issuer authorization policy.

### Implementation

`trusted_key_issuers` now uses an optional mapping on both `VerifyRequest` and `VerificationContext`:

```python
trusted_key_issuers: Optional[dict[str, list[str]]] = Field(default=None)
```

The semantics are:

* `None` means issuer authorization is not configured. Existing callers that omit the field retain the previous behavior.
* `{}` means issuer authorization is explicitly configured but authorizes no signing key for any issuer. Verification therefore fails closed for every signing key.
* A populated mapping authorizes the listed issuer SPIFFE URIs for the corresponding key IDs.

`_signature_key_issuer_mismatch()` now distinguishes `None` from an empty mapping by checking explicitly for `None` before applying the issuer authorization policy.

### Reviewer follow-up

The shared collection-size validator originally called `len(v)` unconditionally. After making `trusted_key_issuers` optional, an explicit `null` value could therefore reach the validator and raise `TypeError`.

The validator now accepts `None` and continues to enforce the existing entry-count limit for configured collections:

```python
if v is None:
    return v
if len(v) > MAX_VERIFY_COLLECTION_ENTRIES:
    raise ValueError(
        f"exceeds {MAX_VERIFY_COLLECTION_ENTRIES}-entry limit"
    )
```

A regression test verifies that `VerifyRequest(..., trusted_key_issuers=None)` is accepted and preserves `None`.

### Verification

The change was checked at the model, verification, HTTP, and collection-boundary levels.

* `python/tests/test_verify_post_bounds.py`: **33 passed**
* `python/tests/test_verify.py`: **93 passed, 4 skipped**
* `python/tests/test_am_verify.py`: **55 passed**
* Targeted `trusted_key_issuers` HTTP test: **1 passed**
* `git diff --check`: **clean**

The current GitHub CI suite passes across the supported matrix. The full Python 3.11 Ubuntu test job reports **1606 passed, 10 skipped, 2 warnings**, with 90.23% coverage.

The existing COSE issuer-authorization tests were also reviewed to confirm that authorized issuers remain valid, unauthorized issuers remain mismatches, and hybrid signatures require authorization for every signing key.

The `None` value was additionally traced through request parsing into `VerificationContext` to confirm that it is preserved and is not normalized back to `{}` before reaching the issuer authorization check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -725,7 +725,7 @@ Makes agent-manifest the canonical hardware-verification library for the org: SE
 
 ### Security
 
-**\[SDK\]** Verification can now bind trusted signing keys to authorized issuers. `VerificationContext.trusted_key_issuers` maps each trusted `key_id` to the issuer SPIFFE URIs allowed to sign with it; when supplied, a manifest whose signing key is not authorized for its declared `issuer` is rejected (fail-closed). Opt-in and backward compatible: an empty map preserves prior behavior.
+**\[SDK\]** Verification can now bind trusted signing keys to authorized issuers. `VerificationContext.trusted_key_issuers` maps each trusted `key_id` to the issuer SPIFFE URIs allowed to sign with it; when supplied, a manifest whose signing key is not authorized for its declared `issuer` is rejected (fail-closed). Opt-in and backward compatible: an omitted value preserves prior behavior, while an explicitly empty map authorizes no issuers.
 
 ### Added
 
--- a/python/src/agent_manifest/_verify.py
+++ b/python/src/agent_manifest/_verify.py
@@ -283,6 +283,8 @@ class VerifyRequest(BaseModel):
     )
     @classmethod
     def _cap_collection_size(cls, v: Any) -> Any:
+        if v is None:
+            return v
         if len(v) > MAX_VERIFY_COLLECTION_ENTRIES:
             raise ValueError(
                 f"exceeds {MAX_VERIFY_COLLECTION_ENTRIES}-entry limit"
--- a/python/src/agent_manifest/_verify.py
+++ b/python/src/agent_manifest/_verify.py
@@ -239,7 +239,7 @@ class VerifyRequest(BaseModel):
     # key_id (sha256 hex of pub key bytes) -> base64url-encoded public key bytes
     trusted_keys: dict[str, str] = Field(default_factory=dict)
     # key_id -> issuer SPIFFE URIs authorized to use that signing key
-    trusted_key_issuers: dict[str, list[str]] = Field(default_factory=dict)
+    trusted_key_issuers: Optional[dict[str, list[str]]] = Field(default=None)
     # principal_id -> base64url-encoded public key bytes (for delegation chain)
     delegation_public_keys: dict[str, str] = Field(default_factory=dict)
     # approver_id -> base64url-encoded Ed25519 public key bytes (for HITL)
@@ -355,7 +355,7 @@ class VerificationContext(BaseModel):
     # key_id (sha256 hex of pub key bytes) -> base64url-encoded public key bytes
     trusted_keys: dict[str, str] = Field(default_factory=dict)
     # key_id -> issuer SPIFFE URIs authorized to use that signing key
-    trusted_key_issuers: dict[str, list[str]] = Field(default_factory=dict)
+    trusted_key_issuers: Optional[dict[str, list[str]]] = Field(default=None)
     # principal_id -> base64url-encoded public key bytes (for delegation chain)
     delegation_public_keys: dict[str, str] = Field(default_factory=dict)
     # approver_id -> base64url-encoded Ed25519 public key bytes (for HITL)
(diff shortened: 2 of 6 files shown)


rajnisht7:
Explicit null is still crashing, PR makes `Optional[...] = None` but validator still do `len(v)`
```python
@field_validator(..., "trusted_key_issuers", ...)
def _cap_collection_size(cls, v):
    if len(v) > MAX_VERIFY_COLLECTION_ENTRIES:  #  TypeError on None
```
this can be fixed with
```python
if v is None:
    return v
if len(v) > MAX_VERIFY_COLLECTION_ENTRIES:
```

litzki-systems:
Thanks, @rajnisht7, for catching this. Fixed the validator to accept `None` for the now-optional `trusted_key_issuers` field while keeping the existing collection-size check for configured mappings.

I also added a regression test and reran the relevant verification and HTTP tests. All passed.

imran-siddique (approved):
Approving. This is the contract from #413 with no gap in it, and I tested the semantics rather than reading them:

```
accepted   None: not configured
MISMATCH   {}: configured, authorizes nothing
accepted   populated, key authorized
MISMATCH   populated, wrong issuer
MISMATCH   populated, different key_id
```

The last two are what make the middle one mean something. A change that only demonstrated `{}` failing closed would be satisfied by an implementation that refuses everything.

@rajnisht7's finding is the one I would not have caught from the diff. `_cap_collection_size` still called `len(v)`, so an explicit `None` raised `TypeError` from the field validator before any of the new semantics ran, which means the `Optional` annotation alone would have made the documented case worse rather than better. Fixed at `3529f6e` with the `v is None` guard, and the size check still applies to a configured mapping.

`AM-VEC-024` lands as you first wrote it in #413, which is the right outcome: the vector was correct and the contract was not.

The suite is 1605 passed and 10 skipped here. The single local failure, `prometheus_client` missing in the docs tutorial test, fails identically on a clean `main` worktree and is my environment.

One sequencing note, not yours to fix. #373 also appends to `CHANGELOG.md`, so merging this conflicts it. That is the structure of the file rather than anything either of you did, and I am handling it on that thread.

Merging.

Issue #340, opened 2026-08-24
Define MCP discovery artifacts and dynamic tool-catalog checkpoints


imran-siddique:
Clarify how Agent Manifest binds an MCP server identity, discovery endpoint, protocol version, and approved catalog-checkpoint policy. Clarify that tool_manifest.catalog_hash can pin a versioned discovery checkpoint rather than assuming a permanently static complete catalog. Keep workload identity separate from OAuth or cA2A delegated authority. Any normative change must follow the repository RFC process.

imran-siddique:
Keep the discovery/catalog commitment scope coordinated with agentrust-io/cmcp#566 and identity separation with agentrust-io/cmcp#568. The next deliverable is the bounded design and acceptance matrix, before normative changes.

solloek369-arch:
For the bounded design and acceptance matrix requested here, coordinated with agentrust-io/cmcp#566, I think three manifest-side points need to be pinned before “catalog checkpoint” has one reproducible cross-repository meaning:

1. **Checkpoint identity and internal consistency.** At current Agent Manifest `0e729e4`, §3.2.3 defines `catalog_hash` as an RFC 9162 Merkle root over `tool_id`, `schema_hash`, and `description_hash`; current cMCP `5b63642` uses the same name for SHA-256 over the full sorted catalog-entry array. Even their empty-catalog values differ. Should these remain deliberately distinct commitments with an explicit projection/join, or should one become the shared checkpoint? Separately, `ToolManifestBinding` accepts a declared root beside `tools[]` without recomputing it; one deliberately wrong-root case would make root↔leaf consistency explicit.

2. **Existing transition evidence.** The cMCP implementation from agentrust-io/cmcp#517 and agentrust-io/cmcp#519 already supplies a detached signed catalog-change record binding previous/new catalog hashes and the change set, with `new_catalog_hash` checked against the runtime catalog. Its documentation says the record chain is not a freshness oracle: records remain replayable after their approvals expire, while an expected predecessor checkpoint must come from an external pin or transparency receipt. It therefore supplies transition provenance, not standalone current authorization. If the AM and cMCP commitments remain distinct, what exact projection or relation should bind that transition evidence to the checkpoint defined here? I am asking for that join, not proposing a second approval mechanism or treating historical provenance as live authority.

[...]

PR #418, opened 2026-09-13
fix(verify): recompute the declared tool catalog root

### LINKED ISSUE #416: Verification never recomputes catalog_hash from tools[], so the §3.2.3 Merkle root is unenforced
`ToolManifestBinding` carries a declared `catalog_hash` beside the `tools[]` it commits to, and verification never recomputes one from the other. Raised by @solloek369-arch on #340 and verified on `main`.

## The gap

`python/src/agent_manifest/models.py:500`:

```python
class ToolManifestBinding(SpecModel):
    """Artifact #3 - spec Section 3.2.3."""
    catalog_hash: HashValue
    tools: list[ToolEntry]
```

`_merkle.py:382` has the builder that returns the `catalog_hash` for a tool manifest. It is imported by `_audit_continuity.py`, `_memory_delta.py` and `_plugins.py`. It is **not** imported by `_verify.py`.

What verification does instead, `_verify.py:1086`:

```python
tm = artifacts.get("tool_manifest") or {}
fields.tool_manifest = _check(
    "tool_manifest",
    tm.get("catalog_hash"),
    context.tool_catalog_hash,
)
```

The declared root is compared against a value supplied in the verification context. The `tools[]` array sitting in the same object is never hashed.

## Why this matters

§3.2.3 defines `catalog_hash` as an RFC 9162 Merkle root over per-tool leaves and says the construction "ensures that any mutation of either a tool's schema or its description, the primary MCP tool poisoning attack vector, breaks the `catalog_hash` Merkle root".

That holds only if something recomputes the root. On the verification path nothing does, so a manifest may declare any root over any `tools[]` and pass. The leaves and the root are two independent assertions by the same producer, and the tree provides no assurance between them.

This is not a weakness in the construction. The construction is correct and the spec text is right about what it would establish. It is simply not being evaluated where it matters.

## Fix

1. Recompute the Merkle root from `tools[]` during verification and refuse a manifest whose declared `catalog_hash` does not match. This is the enforcing control.
2. Keep the existing context comparison. It answers a different question, whether the approved catalog is the one the relying party expects, and both are needed.
3. Add a deliberately wrong-root vector: same `tools[]`, a root that does not correspond, expected result MISMATCH. Without it a future refactor can drop the recomputation and every existing test still passes, which is how this arose.

Point 3 is the one that keeps it fixed.

## Scope

Verification only. No change to the construction in §3.2.3, no change to the wire shape, no change to how `context.tool_catalog_hash` is supplied.

Related: #340, agentrust-io/cmcp#566.

noah-ing:
## What

Recompute a supplied tool catalog's declared Merkle root during `verify_manifest()`, independently of the existing runtime-context comparison.

## Why

Closes #416. This implements the issue reported by @solloek369-arch on #340 and confirmed/filed by @imran-siddique; I am not claiming the original discovery.

A valid signature authenticates the declared root and tools, but does not make them consistent. Previously, an incorrect root could return `VALID` when the caller supplied that same root in `VerificationContext`. The initial regression run on unchanged production code produced **12 failures / 15 passing controls**.

The patch uses the existing `build_catalog_tree()` with the root's declared algorithm. Internal mismatches are reported as `tool_manifest.catalog_hash`; the existing `tool_manifest` runtime comparison remains independent. Neither can override the other into a match. Missing committed leaf inputs and unusable Python-dict inputs fail with a bounded diagnostic.

Compatibility is deliberate: legacy omitted bindings retain their existing behavior; an explicitly empty `tools` array is hashed. Nonhashed metadata does not become a new required input. No Merkle construction or public return shape changes.

## Spec impact

None; SDK enforcement of the existing §3.2.3 construction. This does **not** unify the Agent Manifest Merkle root with cMCP's separate catalog commitment (#340), authenticate live tool schemas from their declared hashes, or establish hardware/runtime behavior.

## Test plan

- [x] Full host suite: **1,647 passed / 10 existing optional skips**.
- [x] Clean committed-source, no-cache Linux container: **1,647 passed / 10 existing optional skips**, 90.26% coverage; runtime network disabled.
- [x] `mypy src/agent_manifest`: 26 files clean.
- [x] Exact CI lint command: `ruff check src/ tests/ --select E,F,W --ignore E501`.
- [x] **41 added cases**: signed v0.1 JSON and v0.2 COSE; SHA-256/SHAKE256; wrong roots; each committed leaf mutation; reordered/empty catalogs; separate runtime mismatch; missing runtime context; legacy omissions; malformed Python-dict inputs. No new skip/xfail.
- [x] Bandit, dependency audit (`--skip-editable`), Gitleaks diff scan, detect-secrets changed-file scan, and whitespace checks pass.
- [x] Wheel/sdist build and Twine validation; each artifact installed into a separate fresh environment, imported and verified outside the checkout, with CLI smoke and archive inspection.
- [x] Changelog updated with attribution. A separate source-review pass found a Python-dict coercion edge case, now rejected and regression-tested.

Both host and container verification use the repository's hash-locked `requirements/dev.txt`; no dependencies were changed. Optional skips are reported, not counted as passes. No live TPM/TEE claim is made by these tests.

## DCO

All commits are signed off as Noah Ingwers (`git commit -s`). By submitting this PR I certify the [Developer Certificate of Origin](https://developercertificate.org/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,14 @@
 
 ### Fixed
 
+- **[SDK]** `verify_manifest()` now recomputes a tool catalog's Merkle root
+  from its supplied `tools` before accepting the declared `catalog_hash`.
+  A valid signature and a matching runtime hash no longer hide inconsistent
+  catalog contents (spec 3.2.3, issue #416). The runtime comparison remains
+  independent, and legacy bindings that omit `tools` retain their existing
+  hash-comparison behavior. Observed by solloek369-arch on #340 and filed as
+  #416 by Imran Siddique.
+
 - **[SDK]** `verify_manifest()` returned `MISMATCH`/`EXPIRED`/`INVALID`/
   `UNVERIFIABLE` for `hitl_record` when *any* approval in `hitl_record.approvals`
   failed, even if a later approval in the same array was present, valid,
--- a/python/src/agent_manifest/_verify.py
+++ b/python/src/agent_manifest/_verify.py
@@ -1088,6 +1088,39 @@ def verify_manifest(
         tm.get("catalog_hash"),
         context.tool_catalog_hash,
     )
+    # Spec 3.2.3: the declared root also commits to the adjacent tools. A
+    # matching runtime hash cannot establish that internal consistency.
+    # Preserve legacy omitted bindings, but hash an explicitly empty catalog.
+    if tm.get("catalog_hash") is not None and "tools" in tm:
+        from ._merkle import build_catalog_tree
+        from ._types import HashValue
+        from .models import ToolEntry
+
+        try:
+            # HashValue fields were validated by the schema gate, but its str
+            # coercions are not applied to this raw dict. Reject Python-only
+            # tool IDs before passing them to the builder.
+            catalog_tools = []
+            for tool in tm["tools"]:
+                if not isinstance(tool, dict) or not isinstance(tool.get("tool_id"), str):
+                    raise ValueError("catalog tools require string tool IDs")
+                # Construct only the Merkle inputs so legacy omissions of
+                # nonhashed metadata do not become new validation failures.
+                catalog_tools.append(ToolEntry.model_construct(
+                    tool_id=tool["tool_id"],
+                    schema_hash=HashValue(tool["schema_hash"]),
+                    description_hash=HashValue(tool["description_hash"]),
+                ))
+            computed_catalog_hash = build_catalog_tree(
+                catalog_tools, algorithm=HashValue(tm["catalog_hash"]).algorithm,
+            )
+        except (KeyError, ValueError):
+            # Missing leaf inputs and construction failures cannot prove a root.
+            computed_catalog_hash = "<unverifiable tools>"
+        if _check(
+            "tool_manifest.catalog_hash", tm["catalog_hash"], computed_catalog_hash,
+        ) == FieldResult.MISMATCH:
+            fields.tool_manifest = FieldResult.MISMATCH
 
     mi = artifacts.get("model_identity") or {}
     # For api-deployed models, bind by version string, not binary hash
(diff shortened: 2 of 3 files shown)


carloshvp (approved):
Reviewed exact head 776677eb6fdbfa02b4634a03d69a14a1aa069091. No blocking findings in this change.

The verifier independently rebuilds the declared catalog commitment from its three leaf inputs while preserving the runtime-context comparison. Signed inconsistent catalogs cannot become valid by supplying the same incorrect root as runtime context. Empty catalogs, both supported hash algorithms, JSON/COSE envelopes, missing leaf inputs, and independent internal/runtime mismatches are covered.

Independent validation in an isolated checkout on macOS/Python 3.13: full suite 1,647 passed, 10 skipped; Ruff's CI selection and mypy passed; whitespace check clean. Against unchanged main verification code, the catalog-focused tests produce 24 failures and 18 passes, confirming regression sensitivity. An additional raw-input matrix returned non-VALID for 42 malformed cases. Three surrogate-containing hash inputs raise during the pre-existing schema gate on both main and this head; that is not introduced by this patch.

Merge simulation against fetched main is conflict-free and the commit has a DCO sign-off. This approval covers the catalog-consistency change on this exact head; it does not attest live schemas or hardware behavior. Remaining repository protection gates are separate from this review.

imran-siddique (approved):
Reviewed 776677eb6fdbfa02b4634a03d69a14a1aa069091.

The catalog root is recomputed from the committed leaf inputs independently of the runtime hash comparison. Empty catalogs, both hash algorithms, JSON/COSE envelopes, malformed leaf inputs, and legacy omissions are covered. Trial merge with current main is clean. Local validation: 1,647 passed, 10 skipped; CI lint selection and mypy pass. Disabling the root-recomputation guard makes all eight targeted false-root cases fail; restoring it makes them pass. No blocking findings.

PR #435, opened 2026-09-15
fix(sdk): make certificate notAfter boundary inclusive per RFC 5280

rajnisht7:
## What

`check_validity_period()` now treats a certificate's `notAfter` instant as valid (`<=`) instead of rejecting it (`<`).

## Why

RFC 5280 §4.1.2.5 defines the validity period as "notBefore through notAfter, **inclusive**," and `cryptography`'s `not_valid_before_utc`/`not_valid_after_utc` are documented with the same inclusive semantics. The check used a strict `<` on the upper bound, so a certificate was rejected during the exact second of its own `notAfter` timestamp.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO


imran-siddique (approved):
Reviewed the inclusive certificate validity boundary and its callers. The change accepts exactly notAfter while retaining rejection after it; the lower bound stays inclusive. Local certificate-chain suite: 24 passed, including both endpoint controls and rejection one second after expiry. Hosted technical checks pass.

PR #436, opened 2026-09-15
fix(cli): stop overwriting signed_at after signing

rajnisht7:
## What

`manifest sign` was overwriting `signed_at` right after the signer set it, with a differently formatted timestamp (`+00:00` + microseconds instead of `Z` + second precision).

## Why

`Ed25519Signer.sign()` already sets `signed_at` correctly via `_signed_at_now()` the same helper `MlDsa65Signer` and `HybridSigner` use, and the format is regression-tested elsewhere. The CLI had one extra line that immediately overwrote that value with `datetime.now(timezone.utc).isoformat()`, so CLI-signed manifests ended up in a different format than every other signing path.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO


imran-siddique (approved):
Reviewed the CLI timestamp correction. Removing the second assignment preserves the signer's canonical signed_at value; signed_at is outside the signature pre-image, so this fixes metadata formatting rather than changing signature coverage. The sentinel regression checks preservation of the actual signer output. Local CLI/COSE and signing suites: 52 passed. Hosted technical checks pass.

PR #437, opened 2026-09-16
fix(hitl): make duration + timestamp validation fail closed everywhere

rajnisht7:
## What

HITL approval expiry checks were letting things through they should have rejected a missing or zero `approval_duration_seconds` was treated as "never expires", the schema silently accepted numeric strings and booleans as a duration when the actual verifier rejected them, and a timezone-naive `approved_at` crashed the verifier with a `TypeError` instead of failing closed like any other bad approval. This PR fixes all three so duration and timestamp validation stay consistent and fail closed everywhere.

## Why
  

`approval_duration_seconds` bounds how long a HITL approval stays valid (ADR-0006: "prevents permanent blank cheques"), and each bug was a different way that bound could be skipped or the verifier could crash instead of rejecting: `duration=0` made the expiry check trivially true so approvals passed without their signature even being checked, the schema and the shared verifier disagreed on what counts as a valid duration (schema was more permissive), and a naive timestamp slipped past schema and then blew up the comparison against an aware `now`. All three now go through shared, single-source-of-truth helpers so schema, standalone verification, and the integrated verifier can't drift apart on this again.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO


carloshvp (dismissed):
Approved exact head `fefb4a4314d860cf2341fa55935209e3f706aef0`.

The change closes both fail-open/error paths consistently across the Pydantic model, standalone approval verification, and the integrated manifest verifier. I reproduced the baseline behavior on current `main`: a year-old approval with `approval_duration_seconds: 0` was accepted, while a timezone-naive approval raised `TypeError` out of the integrated verifier.

Validation in an isolated macOS/Python 3.11 checkout:

- Full suite: `1795 passed, 10 skipped`.
- Focused HITL/verification/COSE suite: `406 passed, 4 skipped`.
- Independent 27-case matrix covered missing, zero, negative, boolean, numeric-string, fractional, null, collection, oversized, expired, malformed/naive timestamp, tampered-scope, and model/integrated-verifier cases, with a valid signed approval control. Every invalid case failed closed without leaking `TypeError` or `OverflowError`.
- `mypy src/agent_manifest` and the repository's exact Ruff CI command passed.
- `git diff --check` and merge simulation against current `main` passed.

All hosted technical checks are green. The pre-review `Policy: Awaiting maintainer review` failure was a repository gate, not an implementation defect.

imran-siddique (dismissed):
Reviewed fefb4a4314d860cf2341fa55935209e3f706aef0 and Carlos's approval of that same commit. The duration and timestamp guards cover model validation, standalone approval verification, and the integrated verifier; malformed structures and date arithmetic failures fail closed.

Local full suite: 1795 passed, 10 skipped. Hosted technical checks pass. This agrees with Carlos's independently reported baseline reproduction and adversarial checks.

imran-siddique (approved):
Re-reviewed 887fd1c475da5a94747d1c2ed440abfb44ff698a after merging current main. The only manual conflict resolution preserves both changelog entries; the earlier HITL security changes remain intact. The combined full suite passed: 1799 passed, 10 skipped. Hosted technical checks pass. An independent review of this latest push remains required.

carloshvp (approved):
Approved exact head `887fd1c475da5a94747d1c2ed440abfb44ff698a`.

I re-reviewed the merge commit against both parents and the previously approved head. Git remerge analysis shows the only manual conflict resolution was in `CHANGELOG.md`, where both entries were retained; the HITL implementation and its tests are byte-for-byte unchanged from `fefb4a4314d860cf2341fa55935209e3f706aef0`.

Validation in a fresh isolated macOS/Python 3.11 checkout:

- Focused HITL and integrated-verifier tests: `189 passed, 4 skipped`.
- Full suite: `1799 passed, 10 skipped`.
- Repository Ruff and mypy commands passed.
- `git diff --check` passed, and the head already contains current `main` (`53be49ade0e4e230be0f98f5765ce130dd441845`).

All hosted technical and policy checks are green, the PR is mergeable, and there are no unresolved review threads.

PR #440, opened 2026-09-17
docs+tests: document and test the caller-side PLATFORM_INFO composition

rajnisht7:
## What

Shows how to gate `verify_manifest()` on SEV-SNP PLATFORM_INFO policy today, without adding any new API no `VerificationContext` fields, no changes to `verify_manifest()` itself.

## Why

LIMITATIONS.md says `appraise_platform_info()` isn't wired into `verify_manifest()`. True, but not a gap: hardware evidence already reaches the verifier only as a result the caller computed (`VerificationContext.verified_attestation_manifest_hashes`), so a caller can call `appraise_platform_info()` next to `verify_attestation_chain()` and gate on both today. That composition just wasn't documented or tested anywhere. This closes that.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO


imran-siddique (approved):
Approving at `3a5ddd2`. The composition is right: PLATFORM_INFO is appraised only after `verify_attestation_chain()` has passed, and the hash goes into `verified_attestation_manifest_hashes` only when both appraisals do, which `verify_manifest()` already honours with no API change. I pushed one commit restoring punctuation that had dropped out in five places in the tutorial and two in the test docstring; no wording or code changed.

PR #442, opened 2026-09-18
fix(cli): keygen no longer briefly exposes or overwrites the private key

rajnisht7:
## What

`manifest keygen` no longer briefly exposes the private key on disk, and no longer overwrites an existing one.

## Why

The private key was written with the default umask and only chmod'd to 0600 afterward (CRYPTO-008/SEC-005), leaving it briefly world-readable - and on a shared host, another process could open it during that window. Overwriting an existing `private.hex` had a second issue: even with a perfect chmod, anyone already holding the old file open would still read the new key through their existing handle, since permissions don't revoke open file descriptors.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO


carloshvp (approved):
Validated exact head 2f92672b799151c522203307e13e953286569aae.

The private key is created atomically with O_CREAT and O_EXCL at mode 0600, without O_TRUNC, and the final mode is applied through the already-open descriptor. Existing files and symlinks are refused without modification, so neither a pre-opened descriptor nor path substitution can observe replacement key material.

Local validation:
- 37 focused CLI and adversarial tests passed, including transient creation permissions, regular and special existing paths, fchmod failure cleanup, and eight concurrent keygen processes producing exactly one matching keypair
- Full PR-head suite: 1,811 passed, 10 skipped
- Full simulated-merge suite against current main: 1,806 passed, 10 skipped
- Repository Ruff, strict mypy, Bandit, and diff checks passed
- Commit carries the required DCO sign-off

The remaining failed policy check is the expected awaiting-maintainer-review gate, not an implementation failure. Approved.

zohebk8s (approved):

imran-siddique (approved):
Approving at `2f92672` as the `python/` code owner. The key is created with `O_CREAT|O_EXCL` at 0600 and the mode is fixed through the open descriptor, an existing file or symlink is refused untouched, and `public.hex` is written only after the private key succeeds, so a refusal cannot leave a mismatched pair.

PR #443, opened 2026-09-19
docs(attestation): note that SNP and TDX providers bind in REPORT_DATA, not the 3.3.1 launch profile

imran-siddique:
The spec and the hardware tutorial disagreed on where the manifest digest sits in a quote. Spec section 3.3.1 says SNP `HOST_DATA`, a TDX `RTMR[3]` extend, and Nitro PCR15. The tutorial says `REPORT_DATA` for SNP and `REPORTDATA` for TDX.

The tutorial matches the code. `SEVSNPProvider` and `TDXProvider` both call `_tsm_get_report(digest + bytes(32))` (python/src/agent_manifest/_hw_providers.py:160 and :582), so the first 32 bytes of the guest-controlled field carry `sha256(pre_image)`. The check reads it back from `report_data[:64 hex]` (python/src/agent_manifest/_attestation.py:368). Nothing in the SDK writes `HOST_DATA` or extends an RTMR. Parsing the two GCP TDX captures in python/tests/fixtures/hardware/gcp-tdx-2026-07-21/ shows a digest in `REPORTDATA[:32]`, zeros in `REPORTDATA[32:]`, and `RTMR[3]` all zeros.

Section 3.3.2 does not reconcile them: its `context_hash` covers policy bundle, system prompt and tool catalog hashes, not the manifest.

So the spec describes a launch-time binding the SDK does not implement yet. I left the normative text alone, since it is consistent with itself. This PR adds one marked note to docs/tutorials/hardware-attestation.md that says what the providers do today, points at section 3.3.1 as the target, and warns that a verifier expecting `HOST_DATA` or `RTMR[3]` will not find the digest there. Docs only. `mkdocs build` shows no new warnings; the 7 strict-mode warnings are pre-existing in changelog.md.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rajnisht7 (approved):
Note correctly documents the gap: the direct SNP/TDX providers bind via guest-controlled REPORT_DATA/REPORTDATA (matches code + the GCP TDX fixtures), while §3.3.1 is the launch-time target the providers don’t implement yet. §3.3.2 callout is accurate too.

PR #444, opened 2026-09-19
fix(cose): make attach_approvals() append instead of replace

rajnisht7:
## What

`attach_approvals()` now adds to the existing approvals instead of replacing them, like `attach_receipt()` already does.

## Why

Attaching approvals one approver at a time silently lost data:
 
```python
env = attach_approvals(env, [alice_approval])
env = attach_approvals(env, [bob_approval])   # Alice's approval was dropped
```
 
ADR-0006 says approvals are collected from several approvers after signing, and the verifier only needs one valid approval (spec 5.3). So attaching one at a time should work. A relying party that trusts only Alice's key went from `APPROVED` to `UNVERIFIABLE` as soon as Bob attached.
 
This is not a signature bypass but it leads to silent data loss.
 
Other changes:
- `approvals` must be a list, otherwise `CoseStructureError`.
- If the envelope already has an approvals value that is not an array (`null` included), it raises `CoseStructureError` instead of overwriting it.
- To replace the whole array, use `attach_unprotected(env, LABEL_APPROVALS, approvals)`.
- The COSE fuzz target now also covers `attach_approvals()`.
Behaviour change: code that relied on the second call replacing the first must switch to `attach_unprotected()`.

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO


imran-siddique (approved):
Approving as the `python/` code owner. The loss was real and the tests bind the fix: on this head the COSE suite passes 146, and with main's `_cose.py` restored 21 of them fail. Refusing a non-array value rather than overwriting it is the right default, and ADR-0006 is the reason one-at-a-time attachment has to work.

PR #445, opened 2026-09-20
fix(tdx): check the PCK chain with the shared verify_cert_chain

rajnisht7:
## What

`verify_tdx_quote()` now checks the PCK chain with the shared `verify_cert_chain()`, like the TPM path does. Malformed chain data is now a `TdxVerificationError` / `CertChainError` instead of a raw exception.

## Why

Step 4 only checked signatures, the root pin and validity. It skipped issuer/subject names, `BasicConstraints` CA, `pathLenConstraint` and `KeyUsage` `keyCertSign`. So a chain with a non-CA intermediate or broken name chaining still verified, verified and reproduced all of these.

Malformed input also escaped as a raw `ValueError` or `UnsupportedAlgorithm` (corrupt PEM, off-curve attestation key, bad root, malformed extensions, unsupported issuer key). The PEM inside a quote is unsigned until it is chained, so an untrusted quote could crash `verify_attestation_chain()` instead of returning `FAILED`. The TPM path had the same escapes through `verify_cert_chain()`.

Real Intel chains pass the stricter checks, and the GCP captures still verify.

Behaviour change: a custom `trusted_root_pem` must now be a CA certificate (`BasicConstraints` `CA=TRUE`).

## Spec impact

None

## Test plan

- [x] `pytest -v` passes

## DCO


carloshvp (dismissed):
Validated exact head 32d1f730d8e843b2aa6f99d9de2e7548a9b8bc18. This closes a real fail-open chain-appraisal gap: the current main branch accepts otherwise-valid TDX quotes whose PCK chain has a non-CA intermediate, a violated path-length constraint, key usage that forbids certificate signing, or broken issuer/subject name linkage; the same probes reject on this head. Malformed PEM, malformed extensions, and unsupported keys are normalized to the verifier error contract, while the genuine Intel captures continue to pass.

Local validation: 1,855 passed and 10 skipped on the pinned cryptography version; the focused certificate-chain, TDX, and TPM set passed 133 tests with one optional skip. Ruff, strict mypy, diff check, DCO, and a merge simulation against current main are clean. CI is green apart from the expected maintainer-review policy gate. Approved.

imran-siddique (dismissed):
Approving as the `python/` code owner, at the merge commit I just pushed. The gap was fail-open on untrusted data and the tests distinguish it: 20 of the new cases fail against main's `_tdx_verify.py` and pass here, the real GCP captures still verify, and malformed PEM, an off-curve attestation key and a bad root now return the verifier's error contract instead of escaping. Using the same `verify_cert_chain()` as the TPM path is the part that keeps this from drifting again.

The `CHANGELOG.md` conflict came from merging #444 ahead of it, so I resolved it with both entries kept.

